PR #27, opened 2018-06-09
Adding example of loading main.css.pode file

ssugar:
To clarify how to load non-view pode files

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,20 @@ body {
 }
 ```
 
+To load the above `.css.pode` file:
+
+```html
+<!-- /views/index.pode -->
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -432,6 +432,20 @@ body {
 }
 ```
 
+To load the above `.css.pode` file:
+
+```html
+<!-- /views/index.pode -->
+<html>
+   
+   <link rel="stylesheet" href="styles/main.css.pode">
```
Badgerati: The `<link />` tag needs to be wrapped in `<head> ... </head>`

Issue #386, opened 2019-09-27
Add a -ContentType parameter to Set-PodeResponseAttachment


Badgerati:
At the moment, the `Set-PodeResponseAttachment` function assumes the content type of the response based on the attachment/file's extension (ie: `.json` gets mapped to `application/json`).

In most cases the extensive mapping is correct. However, there should be a new `-ContentType` parameter on the `Set-PodeResponseAttachment` function that, if supplied, will be used instead of the mapping.

This new parameter should be an optional string parameter, with the help comment updated. If supplied, the `$WebEvent.Response.ContentType` should be set as this value instead of:

```powershell
$WebEvent.Response.ContentType = (Get-PodeContentType -Extension $ext)
```

Badgerati:
Completed by #388

PR #388, opened 2019-10-01
Add -ContentType Parameter, helps #386

Windos:
### Description of the Change

This PR adds a `ContentType` parameter to `Set-PodeResponseAttachment`.

I initially wanted to validate the input based on a full set of valid options... but there's way more than is tidy/practical to include.

Instead I've used a regex pattern to make sure it matches the format of "category/type". When this check fails the output isn't pretty, a follow up PR could be to move this over to ValidateScript with a nicer custom error message.

### Related Issue

#386 

### Examples

```powershell
Set-PodeResponseAttachment -Path './data.txt' -ContentType 'application/json'
```

This command specifies that the content stored in data.txt should be considered as JSON data rather than plain text.

### DIFF
--- a/src/Public/Responses.ps1
+++ b/src/Public/Responses.ps1
@@ -37,7 +37,7 @@ function Set-PodeResponseAttachment
         [string]
         $Path,
 
-        [ValidatePattern('\w+\/[\w.-]+')]
+        [ValidatePattern('^\w+\/[\w\.\+-]+$')]
         [string]
         $ContentType
     )
--- a/src/Public/Responses.ps1
+++ b/src/Public/Responses.ps1
@@ -26,8 +26,6 @@ Set-PodeResponseAttachment -Path 'c:/content/accounts.xlsx'
 
 .EXAMPLE
 Set-PodeResponseAttachment -Path './data.txt' -ContentType 'application/json'
-
-This command specifies that the content stored in data.txt should be considered as JSON data rather than plain text.
 #>
 function Set-PodeResponseAttachment
 {
--- a/src/Public/Responses.ps1
+++ b/src/Public/Responses.ps1
@@ -11,6 +11,10 @@ The Path to a static file relative to the "/public" directory, or a static Route
 If the supplied Path doesn't match any custom static Route, then Pode will look in the "/public" directory.
 Failing this, if the file path exists as a literal/relative file, then this file is used as a fall back.
 
+.PARAMETER ContentType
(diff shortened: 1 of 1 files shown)


src/Public/Responses.ps1
```diff
@@ -19,14 +23,23 @@ Set-PodeResponseAttachment -Path './image.png'
...
 function Set-PodeResponseAttachment
 {
     [CmdletBinding()]
     param (
         [Parameter(Mandatory=$true)]
         [string]
-        $Path
+        $Path,
+
+        [ValidatePattern('\w+\/[\w.-]+')]
```
Badgerati: Some valid content types have a `+`, such as: `application/windows-search-connector+xml`. There's a massive list of content types within the `Private/Mappers.ps1` file. (when I added the `+` it appeared to match them all - I'd also advise escaping the `.` in case it starts matching every character on some platforms).

Ah, I'd also recommend adding `^` to the start, and `$` to the end - to avoid erroneous text around the type.

src/Public/Responses.ps1
```diff
@@ -19,14 +23,23 @@ Set-PodeResponseAttachment -Path './image.png'
 
 .EXAMPLE
 Set-PodeResponseAttachment -Path 'c:/content/accounts.xlsx'
+
+.EXAMPLE
+Set-PodeResponseAttachment -Path './data.txt' -ContentType 'application/json'
+
+This command specifies that the content stored in data.txt should be considered as JSON data rather than plain text.
```
Badgerati: The documentation for Pode's function is auto-generated using PlatyPS, and at the moment, it doesn't play too well with multi-line examples. I'll add this into the contributing file, but just the first line with the code example is all right, I'd remove the descriptive one - so long as the new parameter's description is good enough (which it is 👍).

Badgerati:
There are indeed a lot of content types! Going for the regex was a good idea.

The error message is OK; but agreed it can be changed in a separate PR to something more descriptive.

src/Public/Responses.ps1
Windos: Good catches, thanks. I misses the + when looking through the list (I actually used that mappers list to generate a list for ValidateSet... yeah, way too long.)

Commit incoming.

Windos:
Awesome thanks for the quick review @Badgerati! Changes have been made. Escaped both the `+` and the `.` and added the beginning and end anchors.

Badgerati:
All looks good, I'll get this merged; thanks!

PR #392, opened 2019-10-07
Make functions pode

Fraham:
### Description of the Change
To keep on Pode functions within the namespace of Pode, there are a few functions that don't use the standard of having `Pode` after the verb

        

- `Test-IsUnix`
- `Test-IsWindows`
- `Test-IsPSCore`
- `Test-IsEmpty`

To       

- `Test-PodeIsUnix`
- `Test-PodeIsWindows`
- `Test-PodeIsPSCore`
- `Test-PodeIsEmpty`


Badgerati:
Looks good 👍 I think, just to keep this backwards compatible for now, it would be worth setting some aliases for the functions for their original names (these can be removed come 2.0.0)

I believe you'll also need to add an "AliasesToExport" in the `Pode.psd1` as well 🤔

Issue #425, opened 2019-12-04
Ability to list all of the routes


Fraham:
### Describe the Feature
Ability to list all of the routes

Badgerati:
Re-opening the issue to add an `-EndpointName` filter to the new `Get-PodeRoute` and `Get-PodeStaticRoute` functions (as mentioned in #426).

ArieHein:
https://github.com/Badgerati/Pode/pull/426#issuecomment-592448519

Badgerati:
Added in support to filter by EndpointName:

```powershell
# all routes
Get-PodeRoute

# all routes for method
Get-PodeRoute -Method Get

# all routes for a Path
Get-PodeRoute -Path '/users'

# all routes for an Endpoint
Get-PodeRoute -Endpoint 127.0.0.1:8080

# all routes for an Endpoint by name
Get-PodeRoute -EndpointName Admin
```

If you assign one route to multiple endpoints:
```powershell
Add-PodeRoute -Method Get -Path '/users' -EndpointName User, Admin -ScriptBlock {}
```

then doing `Get-PodeRoute -Path '/users'` will return two routes - one for each endpoint name. I've also changed `EndpointName` on the returned Route to be literally the name of the endpoint that Route is assigned to - not the array of all.

ArieHein:
This will be very beneficial for my unit testing !

PR #426, opened 2019-12-04
Issue 425

Fraham:
### Description of the Change
Adds a new function for routes to get a list of all the configured routes on the server

### Related Issue
#425 

### Examples
Get-PodeRoutes

### DIFF
--- a/src/Public/Routes.ps1
+++ b/src/Public/Routes.ps1
@@ -814,3 +814,42 @@ function Add-PodePage
     # create the route
     Add-PodeRoute -Method Get -Path $_path -Middleware $Middleware -ArgumentList $arg -ScriptBlock $logic
 }
+
+<#
+.SYNOPSIS
+Gets all the routes.
+
+.DESCRIPTION
+Gets all the routes. Returns a list of routes with each element containing a hashtable with the following properties Method, Path, Endpoint, Protocol, Middleware, ContentType, ErrorType, Arguments, Logic
+
+.EXAMPLE
+Get-PodeRoutes
+#>
+
+function Get-PodeRoutes {
+    param ()
+
+    $routes = @()
+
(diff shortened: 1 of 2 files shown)


Badgerati:
Hey @Fraham, just from what we spoke about earlier:

* Calling it `Get-PodeRoute` - make sure to rename the existing private function (maybe to `Find-PodeRoute` instead 🤔 and update references)
* Letting `Get-PodeRoute` return all routes, or be filtered by Route/Method

Badgerati:
Merged with #486

ArieHein:
Remember that you can now have multiple endpoints on same command as was added in 1.4.0 so i assume this will bring all of them across all endpoints ? with a note one what endpoint it belongs to ? parent-child like ?

Badgerati:
The way routes in Pode are treated does mean if you add a `/` route on some `user` and `admin` endpoints then both will be returned as separate routes:

Assuming `user` is 127.0.0.1 and `admin` is 127.0.0.2, then calling `Get-PodeRoute` will return:
```powershell
@(
    @{
        Path = '/'
        EndpointNames = @('user', 'admin')
        Endpoint = 127.0.0.1
        # etc
    },
    @{
        Path = '/'
        EndpointNames = @('user', 'admin')
        Endpoint = 127.0.0.2
        # etc
    }
)
```

There is an `-Endpoint` and `-Protocol` filter currently on `Get-PodeRoute`, but I must admit I did miss an `-EndpointName` filter! I'll re-open #425 and add that in.

ArieHein:
I usually use same IP but different ports

Badgerati:
The Endpoint property is of the format `<ip>:<port>`, so you can pass `-Endpoint 127.0.0.1:8080` or `-Endpoint 127.0.0.1:8081` and the appropriate route will be returned - though it would be easier to be able to supply just the EndpointName.

PR #449, opened 2020-01-02
Update /src/private/Routes.ps1 - issue 446

jhainau:
### Description of the Change
modify /src/private/Routes.ps1 to remove Foreach-Object in favor or Foreach ($x in @()) {script block }. Used 'break' after first valid response instead of Select-Object -First 1.

### Additional Context
per earlier discussion, this is not really necessary, but does remove instances of _forbidden_ commandlets and simplifies a large group of _IF_ statements into a single _Switch_ statement set

### DIFF
--- a/src/Private/Routes.ps1
+++ b/src/Private/Routes.ps1
@@ -344,9 +344,13 @@ function Get-PodeEndpointByName
     }
 
     # ensure it exists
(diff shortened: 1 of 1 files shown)


Badgerati:
Looks good! 😄 I'll get it merged 👍

Issue #465, opened 2020-01-28
Bug in web-rest-openapi-simple.ps1


haidouks:
### Post method example in web-rest-openapi-simple.ps1 has a bug.
While route definition expects a lower case 'userId' property, it is defined uppercase as Open API property. Also OA property 'Name' has no effect on route example. 

### Steps To Reproduce
1. Go to 'examples/web-rest-openapi-simple.ps1
2. Check line number 43,48 and 49

### Expected Behavior
Post sample on Swagger, should set Name and userId properties.

### Platform
 - OS: MacOS
 - Browser Safari
 - Version 6.2.3

Badgerati:
Closing now it's been merged; thanks!

PR #466, opened 2020-01-28
Update web-rest-openapi-simple.ps1 #465

haidouks:
Fix example

### Description of the Change
Property 'userId' in route is now lowercase. Also 'Name' property comes from post data. 

### Related Issue
https://github.com/Badgerati/Pode/issues/465

### DIFF
--- a/examples/web-rest-openapi-simple.ps1
+++ b/examples/web-rest-openapi-simple.ps1
@@ -40,13 +40,13 @@ Start-PodeServer {
 
     Add-PodeRoute -Method Post -Path '/api/users' -ScriptBlock {
         param($e)
-        Write-PodeJsonResponse -Value @{ Name = 'Rick'; UserId = $e.Data.userId }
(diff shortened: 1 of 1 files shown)


examples/web-rest-openapi-simple.ps1
```diff
@@ -40,14 +40,14 @@ Start-PodeServer {
 
     Add-PodeRoute -Method Post -Path '/api/users' -ScriptBlock {
         param($e)
-        Write-PodeJsonResponse -Value @{ Name = 'Rick'; UserId = $e.Data.userId }
+        Write-PodeJsonResponse -Value @{ Name = $e.Data.Name; UserId = $e.Data.userId }
```
Badgerati: Agree with the `$e.Data.Name`.

For the `userId`, this was likely a copy-paste error 🙈 To match the `Name` format this could be best off being renamed to `UserId` - leaving the below OA definition as `UserId` 🤔

haidouks:
👍

PR #535, opened 2020-04-16
Added Support for IIS based Windows Authentication in trusted domains

### LINKED ISSUE #534: IIS Windows Authentication from a foreign (trusted) domain errors out
### Describe the Bug
IIS Windows Authentication does not work with users from a trusted domain.

### Steps To Reproduce
1. Setup Windows Authentication on IIS (doesn't matter if AppPoolIdentity or Domain Account)
2. Try to authenticate using a user from a trusted domain (transitive bi-directional trust)

### Expected Behavior
Users from trusted domains should be able to authenticate just like users from the local domain, in this setup it works for users from the local domain but not for users from the foreign domain.

### Platform
 - OS: Windows Server 2019
 - Browser Independent
 - Version: Latest PS 7.0, latest .NET Core Hosting Pack and Pode 1.7.0

### Additional Context
I am able to see the successful authentication from the foreign user on the Security Event Log, I can also logon to another application hosted on this IIS with Windows Authentication and the foreign user.
However I get an "401 - Failed to retrieve user from authentication token" from Pode.

I can quickly builtup a demo environment so we can debug this if you need.

This is the error I get when authenticating from a foreign user:
```
Date: 2020-04-16 00:23:40
Level: Error
ThreadId: 5
Server: %Servername%
Category: NotSpecified: (:) [Close-PodeDisposable], ExtendedTypeSystemException
Message: The following exception occurred while retrieving member "Dispose": "There is no such object on the server."
StackTrace: at Close-PodeDisposable, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Public\Utilities.ps1: line 90
at <ScriptBlock>, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\Authentication.ps1: line 475
at Invoke-PodeScriptBlock, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Public\Utilities.ps1: line 646
at <ScriptBlock>, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\Authentication.ps1: line 632
at Invoke-PodeScriptBlock, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Public\Utilities.ps1: line 638
at Invoke-PodeMiddleware, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\Middleware.ps1: line 46
at Invoke-PodeSocketHandler, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\PodeServer.ps1: line 233
at Invoke-PodeSocketProcessAccept, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\Sockets.ps1: line 202
at <ScriptBlock>, C:\Program Files\PowerShell\Modules\Pode\1.7.0\Private\Sockets.ps1: line 63
at <ScriptBlock>, <No file>: line 15
```

RobinBeismann:
### Description of the Change
Adds domain and server context to the ADSISearcher so it queries the domain of the user, this way we can handle users from trusted domains.

### Related Issue
Fixes #534 

### Additional Context
Group Support is also given, however it only returns the groups of the user in its respective domain. Querying all domains and resolving foreign security principals takes too long, even with ADSISearcher - I know that from another project.

ADSISearcher sometimes doesn't handle AD Sites and Services properly resulting in using a Domain Controller in another (possibly far away) site.
To handle this it is possible to query the destination DCs latency before. I don't think this should be handled by Pode but instead by the application on it.
Examples can be found in this snippet [Get-ActiveDirectoryWebservices](https://github.com/RobinBeismann/PowerShell-Scripts/blob/master/Scripts/MS-AD/Migrate-AADIdentity.ps1).

### DIFF
--- a/src/Private/Authentication.ps1
+++ b/src/Private/Authentication.ps1
@@ -465,28 +465,34 @@ function Get-PodeAuthWindowsADIISMethod
             if (![string]::IsNullOrWhiteSpace($domain) -and (@('.', $env:COMPUTERNAME) -inotcontains $domain)) {
                 # get the server's fdqn (and name/email)
                 try {
-                    $ad = [adsi]"LDAP://<SID=$($winIdentity.User.Value.ToString())>"
-                    $user.DistinguishedName = @($ad.distinguishedname)[0]
-                    $user.Name = @($ad.name)[0]
-                    $user.Email = @($ad.mail)[0]
+                    $searcher = [adsisearcher]""
+                    $searcher.SearchRoot = [adsi]"LDAP://$($domain)"
+                    $searcher.Filter = "ObjectSid=$($winIdentity.User.Value.ToString())"
+
+                    $ad = $searcher.FindOne()
+                    
+                    $user.DistinguishedName = @($ad.Properties.distinguishedname)[0]
+                    $user.Name = @($ad.Properties.name)[0]
+                    $user.Email = @($ad.Properties.mail)[0]
                     $user.Fqdn = (Get-PodeADServerFromDistinguishedName -DistinguishedName $user.DistinguishedName)
                 }
                 finally {
-                    Close-PodeDisposable -Disposable $ad -Close
+                    Close-PodeDisposable -Disposable $searcher
                 }
 
                 try {
-                    # open a new connection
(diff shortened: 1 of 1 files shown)


Badgerati:
LGTM!

PR #639, opened 2020-12-05
Update web-pages-docker.ps1

ArieHein:
changed port to match the example

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,4 +1,4 @@
 FROM badgerati/pode:test
 COPY . /usr/src/app/
-EXPOSE 10000
+EXPOSE 8085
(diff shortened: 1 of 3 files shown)


Badgerati:
Apologies for the late reply on this one!

The 10000 likely comes from when I had something else running on 8085, but forgot to switch the example back. If we switch the example back to 8085, then the `docker-compose.yml` and `Dockerfile` in the examples directory will also need updating.

Conversely, the docs could be tweaked to be port 10000 instead 🤔 either is fine.

ArieHein:
for commonly-used-ports, I think 8085 would be better. As we tend to associate the 80, 80xy for http related ports.

PR #670, opened 2021-01-21
Remove extra NewLine from form files

ili101:
### Description of the Change
When trying the example code from the issue if you compare the source file to the file downloaded they are identical except the downloaded file have "0D 0A" HEX which is ASCII CRLF added to the end of it.
I found that in the code there was a case to remove it but the length to index was missing the -1 adjustment

### Related Issue
https://github.com/Badgerati/Pode.Web/issues/15


Badgerati:
Tested and looks good, thanks!

PR #693, opened 2021-02-12
Allow OpenApi OperationId to be specified

glatzert:
### Description of the Change
I added a parameter to Set-PodeOARouteInfo to be able to specify an operationId for the route.

### Additional Context
This is relevant, if you use client-gens like nswag to create meaningful names for operations independent of the routes.

### DIFF
--- a/src/Public/OpenApi.ps1
+++ b/src/Public/OpenApi.ps1
@@ -1314,6 +1314,9 @@ A quick Summary of the route.
 .PARAMETER Description
 A longer Description of the route.
 
+.PARAMETER OperationId
+Sets the OperationId of the route.
+
 .PARAMETER Tags
 An array of Tags for the route, mostly for grouping.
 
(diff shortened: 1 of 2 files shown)


src/Public/OpenApi.ps1
```diff
@@ -1343,6 +1343,10 @@ function Set-PodeOARouteInfo
         [string]
         $Description,
 
+        [Parameter()]
+        [string]
+        $OperationId,
```
Badgerati, line 1348: This will need a new `.PARAMETER` comment above the function,

Badgerati (changes requested):

src/Public/OpenApi.ps1
glatzert: I added a short description.

Badgerati (approved):
LGTM

PR #729, opened 2021-04-08
Fix something, i did wrong

glatzert:
Me again.

I did a small error in the component reference (missing `/`).
This adds the missing `/`


Badgerati:
Eek, I missed that as well! 🙈

PR #800, opened 2021-08-04
Update and rename FunctionsAndModules.md to AutomaticFunctionsAndModu…

phatmandrake:
…les.md

I added an example of how to use an automatically generated route. The name of this section should also reflect what it's for.


docs/Tutorials/Routes/Utilities/AutomaticFunctionsAndModules.md
```diff
@@ -28,6 +28,13 @@ Add-PodeRoute -Method Post -Path '/Invoke-Expression' -ScriptBlock {
 }
 ```
 
+Example Route Invocation:
+
+```powershell
+Invoke-Restmethod http://localhost:8080/Invoke-Expression -Body @{Command = "Get-Date"}
+```
+Function parameters become hashtable key value pairs passed in the body of the request.
+
```
Badgerati, line 37: "Invocation" should be a lowercase "i" :)

I actually just spotted that on line 19, `$WebEvent.Data` should be `$WebEvent.Query` for a GET call; as GET is via query string, and any other method is in the body payload. So might be worth showing both:

```powershell
Invoke-RestMethod -Uri 'http://localhost:8080/Invoke-Expression' -Body @{Command = "Get-Date"} -Method Post
Invoke-RestMethod -Uri 'http://localhost:8080/Get-ChildItem?Path=/some/path'
```

Badgerati (changes requested):

Badgerati:
Oh, just noticed, the PR should be against the develop branch.

docs/Tutorials/Routes/Utilities/AutomaticFunctionsAndModules.md
phatmandrake: Updated - also added an additional example for using the -body parameter for query parameters as well.

Badgerati (approved):

Badgerati:
LGTM!

PR #900, opened 2022-01-10
Open/Close RunspacePools async, and load modules in pipeline so BeginInvoke blocks less

### LINKED ISSUE #896: Investigate opening RunspacePools async
Investigate to see if it's possible to start all RunspacePools async, rather than sync. This could speed up the Pode server starting - especially if a lot of modules are being used, or for any particularly slow ones (ie: #892).

We should be able to use `.BeginOpen` and `.EndOpen` on the RunspacePools, and check their StateInfo: Looks like there’s a `BeginOpen` and then a `StateInfo` : [runspacepool docs](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.runspaces.runspacepool?view=powershellsdk-7.0.0#properties)

Plan could be to call BeginOpen (might have to be in the PodeListener, unless we can pass a null AsyncCallback), and then check the StateInfo. If any are Broken, apparently we can call EndOpen to get back any errors.

Badgerati:
### Description of the Change
Hefty change to the way Pode starts up and closes. This makes it so that RunspacePools are now opened/closed asynchronously, and moves module loading to the pipeline level rather than the runspace level. This allows a server to start-up faster, and doesn't block on starting a runspace pipeline.

The "Listening on" message will now appear when all RunspacePools have opened, and at least 1 runspace that runs and endpoint (http, smtp, etc) has started.

Loading the modules in the pipeline works the same as loading them in the runspace - in that once they're imported, re-using a runspace won't re-import everything again, but this way BeginInvoke doesn't forever block on slow loading modules.

### Related Issue
Resolves #896

### DIFF
--- a/examples/server.psd1
+++ b/examples/server.psd1
@@ -43,6 +43,9 @@
             Functions = @{
                 ExportOnly = $true
             }
+            Modules = @{
+                ExportOnly = $true
+            }
         }
         Request = @{
             Timeout = 30
--- a/examples/web-pages.ps1
+++ b/examples/web-pages.ps1
@@ -10,7 +10,7 @@ Import-Module "$($path)/src/Pode.psm1" -Force -ErrorAction Stop
 # Import-Module Pode
 
 # create a server, and start listening on port 8085
-Start-PodeServer -Threads 2 {
+Start-PodeServer -Threads 2 -Verbose {
     # listen on localhost:8085
     Add-PodeEndpoint -Address * -Port 8090 -Protocol Http -Name '8090Address'
     Add-PodeEndpoint -Address * -Port $Port -Protocol Http -Name '8085Address' -RedirectTo '8090Address'
--- a/src/Private/Context.ps1
+++ b/src/Private/Context.ps1
@@ -376,7 +376,7 @@ function New-PodeContext
 function New-PodeRunspaceState
 {
     # create the state, and add the pode module
-    $state = [initialsessionstate]::CreateDefault2()
+    $state = [initialsessionstate]::CreateDefault()
     $state.ImportPSModule($PodeContext.Server.PodeModulePath)
 
     # load the vars into the share state
--- a/src/Private/Context.ps1
+++ b/src/Private/Context.ps1
@@ -357,6 +357,9 @@ function New-PodeContext
         Stop = [ordered]@{}
     }
 
+    # modules
+    $ctx.Server.Modules = @{}
+
     # return the new context
     return $ctx
 }
@@ -364,7 +367,7 @@ function New-PodeContext
 function New-PodeRunspaceState
 {
     # create the state, and add the pode module
-    $state = [initialsessionstate]::CreateDefault()
+    $state = [initialsessionstate]::CreateDefault2()
     $state.ImportPSModule($PodeContext.Server.PodeModulePath)
 
     # load the vars into the share state
@@ -463,7 +466,7 @@ function Import-PodeModulesIntoRunspaceState
 
         # import the module
         $path = Find-PodeModuleFile -Name $module
-        $PodeContext.RunspaceState.ImportPSModule($path)
+        $PodeContext.Server.Modules[$module] = $path
     }
 }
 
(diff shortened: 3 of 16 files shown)


Badgerati:
I've been testing these changes quite a bit, still have a bit more to do. With what the changes are, @robinmalik / @Veers01, would either of you be all right to test these changes on your side? Just to make sure I haven't completely knackered Pode even starting! 😂 

All the changes are .ps1 files, no .cs. If you have the building part working then build away 😃, but if not, you should be able to copy the Private/Public folders from src into a copy of the Pode module from PSGallery.

All going well, you should spot a faster start-up/close-down. I've tried with MicrosoftTeams and while it does "increase" the start-up time, their module's so slow I don't think there's much I can do - I tracked down where the slowness is, and it's trying to dot-source literally hundreds of single files. If I remember, on 1 `-Thread` I think it was 20-80s start-up and with 7 is was ~2-3mins. That's with the MicrosoftTeams module of course, normally it's sub-seconds, hahah.

Oh, on a side-note, I think I've also been able to inadvertently fix the "WARNING" that appear when runspaces are loading modules with non-standard verbs 🤔

I've also added more Verbose lines for when you do `Start-PodeServer -Verbose`, mostly around runspaces starting/closing.

Veers01:
Humm, not sure if it is on my side, but when I replace de .ps1 file, I got this error when I tried to lauch a page:

```Listening on the following 1 endpoint(s) [1 thread(s)]:
        - http://localhost:80/
Date: 2022-01-10 16:36:30
Level: Error
ThreadId: 1
Server: LPADTIADMBD
Category: ObjectNotFound: (Get-PodeWebState:String) [Invoke-PodeScriptBlock], CommandNotFoundException
Message: The term 'Get-PodeWebState' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
StackTrace: at <ScriptBlock>, C:\Program Files\WindowsPowerShell\Modules\Pode.Web\0.7.0\Public\Pages.ps1: line 242
at Invoke-PodeScriptBlock, C:\Program Files\WindowsPowerShell\Modules\Pode\2.5.2\Public\Utilities.ps1: line 697
at <ScriptBlock>, <No file>: line 114

Date: 2022-01-10 16:36:30
Level: Error
ThreadId: 1
Server: LPADTIADMBD
Category: System.Management.Automation
Message: The term 'Get-PodeWebState' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
StackTrace:    at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
   at Invoke-PodeScriptBlock(Closure , FunctionContext )
   at System.Management.Automation.PSScriptCmdlet.RunClause(Action`1 clause, Object dollarUnderbar, Object inputToProcess)
   at System.Management.Automation.CommandProcessorBase.Complete()```

**Comment by robinmalik** [comment]
Sure @Badgerati, I'll try testing tomorrow!

**Comment by robinmalik** [comment]
Ok, if I run an existing project with a few imports (`MicrosoftTeams`, `ExchangeOnlineManagement`, `VMware.VimAutomation.Core`) it all works using Pode 2.5.2. Connecting quickly this evening - perhaps out of peak hours for the UK based DCs. For speeding up the Exchange connection/module import it also uses `Connect-ExchangeOnline -CommandName @('Get-Mailbox', 'Get-EXOMailbox', 'Get-MailboxStatistics', 'Get-EXOMailboxStatistics', 'Get-MailboxDatabase')`

If I force import the module from `Pode\src\Pode.psm1` after checking out `Issue-896`, and try and invoke a route that calls `Get-Mailbox` I get:

```
The term 'Get-Mailbox' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At C:\Squirtle\routes\exol.ps1:7 char:12
+             return (Get-Mailbox -Identity $WebEvent.Parameters['id'])
+                     ~~~~~~~~~~~
at <ScriptBlock>, C:\Squirtle\routes\exol.ps1: line 7
at Invoke-PodeScriptBlock, C:\Pode\src\Public\Utilities.ps1: line 708
at Lock-PodeObject, C:\Pode\src\Public\Utilities.ps1: line 188
at <ScriptBlock>, C:\Squirtle\routes\exol.ps1: line 5
at Invoke-PodeScriptBlock, C:\Pode\src\Public\Utilities.ps1: line 697
at <ScriptBlock>, <No file>: line 114
    + ObjectNotFound: (Get-Mailbox:String) [], CommandNotFoundException
```
This seems odd because a route that calls `Get-EXOMailbox` works just fine and these are commands in the same Exchange module.

Both tests were done using a new PowerShell session (window) running as Administrator, multiple times. I tried visiting the `Get-EXOMailbox` route first - which way round didn't matter.

I'm just going to try removing the limitation on cmdlets to see if that changes anything.

**Comment by Badgerati** [comment]
Hey @robinmalik,

That's bizarre! I think I can see why, the ExchangeOnlineManagement module doesn't actually export the `Get-Mailbox` cmdlet; it seems to be some "special magic" that happens on `Connect-ExchangeOnline` 🤔 It kind of make sense, but also not because I can't think of why the old runspace logic would work, and the new wouldn't, hmm.

The Import logic should be working the same, though if the module doesn't actually export it, then it's a bit baffling! Only idea I have is this line https://github.com/Badgerati/Pode/blob/e220c4918d4d8e276cc808d9b0160f375dd76109/src/Private/Context.ps1#L370

I changed it to `CreateDefault2()`, if you change that to `CreateDefault()` does that work? I don't have an Exchange I can connect to to test with unfortunately 😢

If it helps, here's a fully compiled version 😃  [pode.zip](https://github.com/Badgerati/Pode/files/7849973/pode.zip) @Veers01 
 
Thanks to both of you by the way! 😊

**Comment by Veers01** [comment]
Thank you for the compiled version! But I still get the error that I mention earlier (Message: The term 'Get-PodeWebState' is not recognized as a name of a cmdlet, function, script...)

Just to be sure, I extract your zip in the 2.5.2 directory of pode Module. Is that correct?

**Comment by Badgerati** [comment]
Hey @Veers01,

Managed to reproduce it 😃 The above commit (or this [pode.zip](https://github.com/Badgerati/Pode/files/7857595/pode.zip)) should resolve the issue. It was down to a scope issue when importing the modules.

(you can either extract the contents into the main gallery location, or extract them into a temp folder and `Import-Module ./path/to/Pode.psm1 -Force`)

Intrigued to know if this fixes the `Get-Mailbox` issue as well, @robinmalik  🤔

**Comment by Veers01** [comment]
@Badgerati: I will test that this evening.

FYI, if you want a Microsoft 365 for testing, you can have one free for 90 day (renewable) here: https://developer.microsoft.com/en-us/microsoft-365/dev-program

**Comment by robinmalik** [comment]
@Badgerati Pulled the change and tested - all good now! 🎉

**Comment by Badgerati** [comment]
Excellent! I'll test some more, but all looking good then! 😄 

@Veers01 thanks for the link :) I've noted it down, as that could definitely come in handy!

**Comment by Veers01** [comment]
Hi! I take some time to test the new version. For what I see, I don't have any problem with the microsoftteams module. I do still have some max shell limitation, but this is a Microsoft limit :)

But, I have some strange thing when I load my custom module. with the normal version, I load my module outside pode and I have access to the function everywhere. But with the patch version, I have to load the module in every page if I want to access the function.

Is this could be a side effect?

**Comment by Badgerati** [comment]
Hi @Veers01,

It could be a side affect yes; have you got an example of what you used to be able to do, and what you now have to do? Plus some information on the custom module - is it a standard `.psm1` module, a binary, script, something else? If you import it and run `Get-Module <name>` what's returned?

**Comment by Veers01** [comment]
Here a example. This is my MFA.psm1 module (enable or disable MFA function):

```#import-module API-Query
$DataFile = "$PSScriptRoot\..\..\Data\MFADesactive.json"

# Récupération de la liste des comptes avec le MFA désactivé et chargement dans MFADisabledUser

function disable-MFA {
	param([string]$userID, [string]$date)

	out-log -level info -message "Desactivation du MFA de $userID jusqu'au $date"

	$MFAUser = Get-PodeState -Name MFADisabledUser
	if($MFAUser.count -eq 0) { $MFAUser = @() }
	$MFAUser = @($MFAUser)

	$MFAUser += @([PSCustomObject]@{ ID = "$userID"; DateTime = "$date" })

	Set-PodeState -Name MFADisabledUser -Value $MFAUser
	Save-PodeState -Include MFADisabledUser -Path $script:DataFile

    $body = @{
        "@odata.id" = "https://graph.microsoft.com/v1.0/directoryObjects/$userID"
    }
    $body = $body | ConvertTo-Json

    $uri = 'https://graph.microsoft.com/v1.0/groups/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/members/$ref'
    out-log -level info -message "Requete API: $uri - $body"
    get-ByAPIQuery -method 'POST' -uri $uri -body $body
}
Export-ModuleMember -Function disable-MFA

function enable-MFA {
	param ([String]$userID)

	out-log -level info -message "Activation du MFA de $($userID)"
    $MFAUser = Get-PodeState -Name MFADisabledUser
	if($MFAUser.count -eq 0) { $MFAUser = @() }
	$MFAUser = @($MFAUser)

	$MFAUser = $MFAUser | where-object { $_.ID -ne $userID }
	$MFAUser = @($MFAUser)
	
	Set-PodeState -Name MFADisabledUser -Value $MFAUser
	Save-PodeState -Include MFADisabledUser -Path $script:DataFile

    $uri = 'https://graph.microsoft.com/v1.0/groups/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/members/{0}/$ref' -f $userID
    out-log -level info -message "Requete API: $uri"
    get-ByAPIQuery -method 'DELETE' -uri $uri
}
Export-ModuleMember -Function enable-MFA
```

Before, I simply load it at the start of the script (outside podestart) and I could use the function anywhere: import-module pathtothemodule/MFA.psm1

now, I have to do the import in each scriptblock when I want to use the function.

```
PS C:\PodeServerX> get-module

ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
Script     0.0                   MFA                                 {disable-MFA, enable-MFA}
```

Veers01:
Ok, after more testing, I dont know where is the bug but I do know this is not a Pode bug :) I revert back to pode 2.5.1 and I still have the error. So, for what I see, everything work find with the change! Thank you very much.

Badgerati:
Hey @Veers01,

I was intrigued and thought I'd test just in case :)

I set up two custom modules in a similar fashion to what you've described, and both of their functions were accessible in routes, pages, timers, etc. The only time I did get "function not found" was when I left `ExportOnly = $true` on in my server.psd1:

```powershell
@{
    Server = @{
        AutoImport = @{
            Modules = @{
                Enable = $true
                ExportOnly = $true
            }
        }
    }
}
```

Setting ExportOnly to false then resolved it. Not sure if you're using a server.psd1, but if so could be worth checking for that value :)

If you need it as true, then you'll need to `Export-PodeModule` in your `Start-PodeServer`.

It's a bit strange, but if it is happening in the previous version then at least it's not caused by this change. If you find the cause and it's Pode, feel free to raise a bug!

Badgerati:
I've a couple more test I want to do for this change, and will merge it soon. But so far seems like it should be safe

Veers01:
> Setting ExportOnly to false then resolved it. Not sure if you're using a server.psd1, but if so could be worth checking for that value :)

You are right, setting it to $false do the trick. I thnik I change it some days ago when I do some testing and forgot to set it back to $false.

The only drawback is that it now take 90 to load the app :(

Badgerati:
That could be modules loaded from your `$PROFILE`.

If you keep it as `$true`, and you know which modules you need, then using `Export-PodeModule` will tell Pode to only import those modules into its runspaces :)

ie:
```powershell
Import-Module MicrosoftTeams
Import-Module path/to/MFA.psm1

Start-PodeModule {
    Export-PodeModule -Name MicrosoftTeams, MFA
}
```

This way, if you have modules like `posh-git` or `terminal-icons`, etc. in your profile, Pode will now ignore them.

Veers01:
That exactly what I have done and this is wayyy more faster! Thank you!

PR #914, opened 2022-01-28
Add async/sync task support

### LINKED ISSUE #893: Add async/sync Task Support
The plan here is to add "Task" support into Pode, which will be scriptblocks that can be invoked adhoc, and will be run async by default but could also be toggled to be sync. This will fix the hack people currently have to do to run a script async, by tricking schedules to "never" run.

For example, you could create some adhoc script, and then trigger it to run async in a route:

```powershell
Start-PodeServer {
    New-PodeTask -Name 'Example' [-ArgumentList @{}] -ScriptBlock {
        # run some task
    }

    Add-PodeRoute -Method Get -Path '/example' -ScriptBlock {
        # run task async
        Invoke-PodeTask -Name 'Example' -[ArgumentList @{}] | Out-Null
    }
}
```

Calling `Invoke-PodeTask` could return a hashtable with a Handle/Result objects, ie: a "task". If you don't care about the result, you could just pipe it to null. But if you do care, then you could pipe the task into a Wait function (after doing other logic in the meantime), and then the Result should be populated if the task `return`s any data:

```powershell
# run the task, wait, then get the result
$task = Invoke-PodeTask -Name 'Example'
# do other things here
$task | Wait-PodeTask
$task.Result
```

Or, if you want the task to be sync anyway:

```powershell
# this will just wait, clean-up, and return any result data by default
$result = Invoke-PodeTask -Name 'Example' -Wait
```

There will be a timer to auto-cleanup old tasks that have completed, but you can manually clean-up by doing the following (`-Wait` will do this by default):

```powershell
$task | Close-PodeTask
```

The tasks will run in their own runspace pool (like schedules), and could have a default max size of 2. Like schedules, this could be customised via some `Set-PodeTaskConcurreny` function.

Badgerati:
### Description of the Change
Adds support for Tasks, which can be run adhoc in either an async or sync manor. Tasks can also return values that can be used.

### Related Issue
Resolves #893 

### Examples
```powershell
Start-PodeServer {
    Add-PodeEndpoint -Address * -Port 8080 -Protocol Http

    Add-PodeTask -Name 'Example1' -ScriptBlock {
        return @{ Value = 'hey look, a value!' }
    }

    Add-PodeTask -Name 'Example2' -ScriptBlock {
        param($value)
        # mimic a long running task
        Start-Sleep -Seconds 10
    }


    Add-PodeRoute -Method Get -Path '/run-task-sync' -ScriptBlock {
        $result = Invoke-PodeTask -Name 'Example1' -Wait
        Write-PodeJsonResponse -Value @{ Result = $result }
    }


    Add-PodeRoute -Method Get -Path '/run-task-sync2' -ScriptBlock {
        $task = Invoke-PodeTask -Name 'Example1'

        # do some other logic here

        $result = ($task | Wait-PodeTask)
        Write-PodeJsonResponse -Value @{ Result = $result }
    }


    Add-PodeRoute -Method Get -Path '/run-task-async' -ScriptBlock {
        # the task will start, and continue running
        Invoke-PodeTask -Name 'Example2' -ArgumentList @{ value = 'custom-values' } | Out-Null
    }
}
```


abhi18av:
Great addition Matthew!

PR #968, opened 2022-05-24
Add support for connecting to external websocket servers

### LINKED ISSUE #895: Investigate the possibility for a server to act as a client for connecting to external WebSockets
Crazy idea here and early days, but what if we could have the Pode server act as a client and connect to other external WebSockets?

So far it seems possible, and it would include something like the following functions:

```powershell
# connect to some external websocket (like slack for example)
Connect-PodeWebSocket -Name 'Slack' -Url '<url>' -ScriptBlock {} [-ArgumentList]

# like Send-PodeSignal, but for external WebSockets
Send-PodeWebSocket -Name 'Slack' -Message 'Message'

# close a current connection
Disconnect-PodeWebSocket -Name 'Slack'

# could be used to reset a connections URL (if it expires). possible? could require current logic?
Reset-PodeWebSocket -Name 'Slack' -ScriptBlock {}
```

The `Connect-PodeWebSocket` would open an async connection to some external websocket endpoint. It will then listen for requests, and call the scriptblock when received. You can send messages to the websocket using `Send-PodeWebSocket`.

This will happen in a seperate runspace, so it can be async. At the moment this runspace will only be created if any `Connect-` calls are made in the main `Start-PodeServer` scope, but I'm wondering if it's possible to create connections adhoc from anywhere 🤔

Badgerati:
### Description of the Change
Adds support for connecting to external websocket servers, for the receiving and sending of messages from your Pode server.

New functions added:
* `Set-PodeWebSocketConcurrency`
* `Connect-PodeWebSocket`
* `Disconnect-PodeWebSocket`
* `Remove-PodeWebSocket`
* `Send-PodeWebSocket`
* `Reset-PodeWebSocket`
* `Test-PodeWebSocket`

### Related Issue
Resolves #895 

### Examples
```powershell
Start-PodeServer -EnablePool WebSockets {

    # listen
    Add-PodeEndpoint -Address * -Port 8092 -Protocol Http

    # connect to web socket
    Connect-PodeWebSocket -Name 'Example' -Url 'ws://localhost:8091' -ScriptBlock {
        Send-PodeWebSocket -Message (@{ message = $WsEvent.Request.Body } | ConvertTo-Json -Compress)
    }

    # connect to web socket via route
    Add-PodeRoute -Method Get -Path '/connect' -ScriptBlock {
        Connect-PodeWebSocket -Name 'Test' -Url 'wss://ws.ifelse.io/' -ScriptBlock {
            $WsEvent.Request | out-default
        }
    }
}
```

### DIFF
--- a/docs/Tutorials/WebSockets/Endpoints.md
+++ b/docs/Tutorials/WebSockets/Endpoints.md
@@ -1,8 +1,6 @@
-# Web Sockets
+# Endpoints
 
-Pode has support for using WebSockets, including secure WebSockets, for either server-to-client or vice-versa.
-
-WebSockets allow you to broadcast messages directly from your server to connected clients. This allows you to get real-time continuous updates on the frontend without having to constantly refresh the page, or by using async javascript.
+Pode has support for creating WebSocket endpoints, for server-to-client and/or client-to-server communications. WebSockets allow you to broadcast messages directly from your server to connected clients. This allows you to get real-time continuous updates on the frontend without having to constantly refresh the page, or by using async javascript.
 
 ## Server Side
 
--- a/docs/Tutorials/WebSockets/External.md
+++ b/docs/Tutorials/WebSockets/External.md
@@ -1,10 +1,10 @@
 # External
 
-Pode has support to connect to external WebSocket servers, and be able to receive messages from then as well as send messages to them as well. This is useful if you want to connect to external metrics servers to update dashboards; send messages to external WebSockets; or even create a bot.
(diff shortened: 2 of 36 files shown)


Badgerati:
Testing this, and in 5.1 the connections can sometimes be forcibly closed - but not in 7+ (like here: https://stackoverflow.com/questions/49178595/net-websocket-closing-for-no-apparent-reason). Think I've found a possible fix using websocket pings.

PR #970, opened 2022-05-30
Add anonymous route access support, when authentication is enabled on a route

### LINKED ISSUE #954: Optional authentication question
### Question
I apologize if I'm just blind and missed this or just didn't understand it in the documentation... Lets say I have a route for login but then I have another route that I want to be able to be called publicly, just do different things if the user already logged into the other page. How do I go about getting what user is logged in?

Badgerati:
### Description of the Change
Add support for anonymous access on routes, plus a new helper function to test whether we have an authenticated user or not.

For the anonymous access, a new `-AllowAnon` switch has been added to `Add-PodeRoute`. If an unauthenticated user hits a Route with this switch then they can access the page - and not be redirected to the FailureUrl/Login page.

For testing if the user accessing the Route is authenticated or not, there's a new `Test-PodeAuthUser` function.

### Related Issue
Resolves #954

### Examples
```powershell
Add-PodeRoute -Method Get -Path '/' -Authentication 'Login' -AllowAnon -ScriptBlock {
    if (Test-PodeAuthUser) {
        $session:Views++
        Write-PodeViewResponse -Path 'auth-home' -Data @{
            Username = $WebEvent.Auth.User.Name
            Views = $session:Views
        }
    }
    else {
        Write-PodeViewResponse -Path 'auth-home-anon'
    }
}
```

Now, when an authenticated user hits the page, they're shown the original personal greeting page with view counter. However, when an unauthenticated user hits the page they are shown a generic greeting with a login button.

### DIFF
--- a/docs/Tutorials/Routes/Examples/AnonymousAccess.md
+++ b/docs/Tutorials/Routes/Examples/AnonymousAccess.md
@@ -1,0 +1,189 @@
+# Anonymous Access
+
+This example builds on top of the [Login Page](../LoginPage) example. This time instead of having a user navigate to the home page, and then be immediately redirected to the login page, this time we'll allow an unauthenticated user to access the home page.
+
+When this unauthenticated user accesses the home page, they'll be a switch in logic to show different content and a login button - rather than a greeting and a logout button.
+
+## Allow Anon
+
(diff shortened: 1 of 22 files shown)


Badgerati:
Oh, this also centralises all of the Using Variable logic for splatting.

Clifra-Jones:
I have tried to use this with Azure AD authentication and I cannot get I to work. I am working with Pode.Web.
I need this for a health check from a load balancer.
 ```
# Health check
    Add-PodeRoute -Method GET -Path '/health' -Authentication 'Login' -AllowAnon -ScriptBlock {
        if (-not (Test-PodeAuthUser)) {
            Write-PodeJsonResponse -Value @{status = 'healthy'}
        }
    }
    
```
Accessing the /health route always present the Microsoft 365 login prompt.

PR #975, opened 2022-06-01
Update readme sample code image to SVG

pcgeek86:
### Description of the Change

Updated README sample code image with a scalable SVG

### Related Issue

n/a


Badgerati (changes requested):
Looks good! Is there any chance the faded background of the image could be made transparent? In dark mode you can see a giant light grey border around the code block 😂

pcgeek86:
@Badgerati I generated this off of Carbon, so I didn't really "create" it. I'd have to figure out how to remove the background, and I don't think that will be easy for an SVG noob like me. 🤣 We can close this out and find a different solution, if you'd like.

Badgerati:
@pcgeek86 was it https://carbon.now.sh/? If so you can set the colour `A` value to `0` to make it transparent 😄

pcgeek86:
@Badgerati can you take another look? I removed it by finding the `<div>` with the [non-existent] CSS className `container-bg` and then deleting it. The shadow is still present, but not the grey background color.

### Before

image


### After 
image

README.md
```diff
@@ -16,7 +16,7 @@
 Pode is a Cross-Platform framework for creating web servers to host [REST APIs](https://badgerati.github.io/Pode/Tutorials/Routes/Overview/), [Web Pages](https://badgerati.github.io/Pode/Tutorials/Routes/Examples/WebPages/), and [SMTP/TCP](https://badgerati.github.io/Pode/Servers/) Servers. Pode also allows you to render dynamic files using [`.pode`](https://badgerati.github.io/Pode/Tutorials/Views/Pode/) files, which are just embedded PowerShell, or other [Third-Party](https://badgerati.github.io/Pode/Tutorials/Views/ThirdParty/) template engines. Plus many more features, including [Azure Functions](https://badgerati.github.io/Pode/Hosting/AzureFunctions/) and [AWS Lambda](https://badgerati.github.io/Pode/Hosting/AwsLambda/) support!
 
 <p align="center">
-    <img src="https://github.com/Badgerati/Pode/blob/develop/images/example_code_2.png?raw=true" />
+    <img src="https://github.com/Badgerati/Pode/blob/develop/images/example_code_readme.svg?raw=true" />
```
Badgerati, line 19: The image looks much better! 😄 

For the `` tag, just so it's scaled appropriately with everything else, adding either `width="70%"` or `width="80%"` will help so it's not overly "large"

That should be good to merge then! :)

Badgerati (changes requested):

README.md
pcgeek86: @Badgerati I've added the `width` attribute in the latest commit. 😄  Thanks for checking it out!

Badgerati (approved):
LGTM

Issue #984, opened 2022-06-08
Bump Docker images to use PS7.2.4


Badgerati:
As per the title.

Issue #974, opened 2022-06-01
Write-PodeTextResponse silently fails if the -Value parameter is not specified


pcgeek86:
### Describe the Bug

Write-PodeTextResponse silently fails if the -Value parameter is not specified.
Pode throws an HTTP error 500.

### Steps To Reproduce

```
Start-PodeServer -ScriptBlock {
 Add-PodeEndpoint -Address 0.0.0.0 -Protocol Http -Port 15002
 New-PodeLoggingMethod -Terminal | Enable-PodeRequestLogging
 Add-PodeRoute -Method Get -Path / -ScriptBlock { Write-PodeTextResponse 'Hello Pode' }
}
```

```
Invoke-WebRequest -Uri http://localhost:15002/
```

### Expected Behavior

If possible, Pode should reject the invalid route configuration, due to the missing parameter.
Errors from the runspace handling the HTTP request should be surfaced to the parent runspace.

### Platform
 - OS: Windows 11
 - Browser: n/a
 - Versions:
   - Pode: 2.5.2
   - PowerShell: 7.2.1

Badgerati:
Hi @pcgeek86,

If you add `New-PodeLoggingMethod -Terminal | Enable-PodeErrorLogging` the error is shown on the terminal :)

Hooowever, it does highlight something that could be added though: if I add the `Position=0` property to the `-Value` parameter then `Write-PodeTextResponse 'Hello Pode'` works.

Issue #965, opened 2022-05-11
Import-module not working in v2.6.2


crackermania:
Hi,

I'm tested v2.6.2 missing capabilities to import-module ps1 file. It's working fine in v2.4.2 and 2.5.2. My code as per below :

```
Import-Module -Name Pode -RequiredVersion 2.6.2

get-module Pode | Out-Default 

#1) not working
#Import-Module "D:\Scripts\testModule.ps1" -Scope Global
#2) not working in 2.6.2, work well in 2.4.2 and 2.5.2
Import-Module "D:\Scripts\testModule.ps1"

Start-PodeServer -Threads 2{

    Add-PodeEndpoint -Address * -Port 8200 -Protocol Http

    Add-PodeRoute -Method Get -Path 'api' -ScriptBlock {		
	#3) not working
        #Import-Module "D:\Scripts\testModule.ps1"
        ("Inside ThreadId: {0}" -f $ThreadId) | Out-Default 	
		$test = randomTaskID
		Write-PodeJsonResponse -Value @{ 'value' = "$test" }
		
    }

    New-PodeLoggingMethod -Terminal | Enable-PodeErrorLogging

}

```
testModule.ps1 contain code as per below : 

```
function randomTaskID
{
	$taskID = Get-Random -Minimum 1000 -Maximum 99999999
	return $taskID
}
```

Consider bug or expected behavior for latest release?

By the way, it's working well if i paste the function before Start-PodeServer.

Thanks.

Badgerati:
Hi @crackermania,

This could be a bug actually for `.ps1` files, the logic for importing modules was tweaked in 2.6.0.

What happens if you change the file to be `.psm1` instead?

crackermania:
@Badgerati ,

Thanks.. I can confirmed, I can use after change to psm1.

From : https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/import-module?view=powershell-7.2

-Name

Specifies the names of the modules to import. Enter the name of the module or the name of a file in the module, such as a .psd1, .psm1, .dll, or .ps1 file. File paths are optional. Wildcard characters aren't permitted. You can also pipe module names and filenames to Import-Module.

So, consider bug? Thanks.

Badgerati:
@crackermania,

Since it used to work pre-2.6.0, and the docs say it should work for ps1 files then yes, it's a bug. Though it seems there's a workaround for now if changing the extension to psm1 works.

Issue #958, opened 2022-04-08
smtp attachment regex not working from scanners


swedan333:
certain scanners include " in boundary data

a regex update might fix the issue

update this line in line 254 in src/Listener/PodeSmtpRequest.cs
match = Regex.Match(line, "^\\s*boundary=(?<boundary>.+?)$");
with this regex to exclude "
^\\s*boundary=("|)(?<boundary>.+?)("|)$

PR #986, opened 2022-06-17
v2.7.0

Badgerati:
### Features
* #895: Add support for server to be able to connect to external WebSockets
* #902: New TCP server, listener, and endpoints; plus Verbs support
* #976: Add support for grouping Routes

### Enhancements
* #901: Enable support for multiple SMTP endpoints, and certificates
* #954: Add anonymous route access support, when authentication is enabled on a route
* #960: Add RSET and NOOP SMTP commands
* #981: Add switch for AccessControl to create global Options route
* #982: Add * level for errors to enable everything

### Bugs
* #956: Fix for importing functionss with inline parameters
* #957: Fix for some OpenAPI properties being dropped
* #958: Some SMTP attachment boundaries can include double quotes
* #965: Fix for importing ps1 files
* #974: Add Position=0 to most response write functions

### Documentation
* #978: Icon and Donation link updates

### Packaging
* #984: Bump Docker images to use PS7.2.4


src/Public/Authentication.ps1
```diff
@@ -1934,4 +1934,15 @@ function ConvertFrom-PodeOIDCDiscovery
         -Middleware $Middleware `
         -CodeChallengeMethod $codeMethod `
         -UsePKCE:$UsePKCE)
+}
+
+function Test-PodeAuthUser
```
Badgerati, line 1939: Needs function summary

src/Public/WebSockets.ps1
```diff
@@ -0,0 +1,238 @@
+using namespace Pode
+
+function Set-PodeWebSocketConcurrency
```
Badgerati, line 3: All functions in file are missing summaries.

Badgerati: added

Issue #1065, opened 2023-01-16
Query parsing error 400


ili101:
### Describe the Bug
Hi
I'm trying to mock a server for an API with Pode.
The client send something like this to Pode:
```powershell
Invoke-WebRequest 'http://example.com/public/v1/assets?eq(status,active)&limit=1&offset=0'
```
Pode returns error 400 and not get into the route (`Add-PodeRoute -Method * -Path * ...`)
Is there a way to process this request?

Note that:
If I send only "limit=1&offset=0" Pode parse the info to `$WebEvent.Query`.
If I send only "eq(status,active)" Pode basically ignores it witch is fine, I can get it form `$WebEvent.Request.Url` if I really need to.
But with both I can't get into a route.

### Expected Behavior
Don't have a strong opinion if "Query" will be set or not, as long as it's available in `$WebEvent.Request.Url` and I can process the call it's enough for me.

PR #1066, opened 2023-01-16
Fix Query with no key parse

ili101:
### Description of the Change
Fix https://github.com/Badgerati/Pode/issues/1065

### DIFF
--- a/src/Private/Helpers.ps1
+++ b/src/Private/Helpers.ps1
@@ -1387,7 +1387,7 @@ function ConvertTo-PodeResponseContent
     if ([string]::IsNullOrWhiteSpace($ContentType)) {
         return ([string]$InputObject)
     }
-    
+
     # run action for the content type
     switch ($ContentType) {
         { $_ -ilike '*/json' } {
@@ -1634,8 +1634,13 @@ function ConvertFrom-PodeNameValueToHashTable
     }
 
     $ht = @{}
+    $i = 0
     foreach ($key in $Collection.Keys) {
-        $ht[$key] = $Collection[$key]
+        if (!$key) {
+            $key = ''
+        }
+        $ht[$key] = $Collection[$i]
+        $i++
     }
 
     return $ht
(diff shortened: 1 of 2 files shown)


ili101:
Test was falsely passing as apparently:
```powershell
$c = [System.Collections.Specialized.NameValueCollection]::new()
$c    $c.Add($null, 'blue')
# Key is ""
```
Is not the same as:
```powershell
$c = [System.Web.HttpUtility]::ParseQueryString('?blue')
# Key is $null
```

`Get-PodeCount` was filtering  '?blue' but not '?blue&foo=bar' which was not consistent, now it will not filter no keys.

src/Private/Helpers.ps1
```diff
@@ -1634,8 +1634,13 @@ function ConvertFrom-PodeNameValueToHashTable
     }
 
     $ht = @{}
+    $i = 0
     foreach ($key in $Collection.Keys) {
-        $ht[$key] = $Collection[$key]
+        if (!$key) {
+            $key = ''
+        }
+        $ht[$key] = $Collection[$i]
```
Badgerati, line 1642: Just in case the Keys and indexes don't line up for some reason, and we fetch the wrong value, would the following be safer by using `.GetValues(...)`? 🤔 

```powershell
$ht = @{}
foreach ($key in $Collection.Keys) {
    $htKey = $key
    if (!$key) {
        $htKey = ''
    }

    $ht[$htKey] = $Collection.GetValues($key)
}
```

Badgerati (changes requested):

src/Private/Helpers.ps1
ili101: LGTM, it works with $null unlike the standard `[$null]`

ili101: Wait it returns arrays now
`{"limit":"1","":"eq(status,active)","offset":"0"}` -> `{"limit":["1"],"":["eq(status,active)"],"offset":["0"]}`

ili101: Updated

Badgerati (approved):
LGTM! :)

PR #1084, opened 2023-03-06
Fix: Test-PodeJwt comparison against Local/Unspecified Kind

### LINKED ISSUE #1083: Test-PodeJwt exp and nbf validation incorrect
### Describe the Bug
`Test-PodeJwt`, responsible for validating JWT payload when using JWT auth scheme will return `The JWT has expired` in certain scenarios based on the machine's local timezone even though the JWT expiry is in future.

### Steps To Reproduce
This can be reproduced by enabling JWT authentication, and changing the machine timezone ahead of UTC.

### Expected Behavior
If JWT exp is in future, Pode should not return `The JWT has expired`

### Platform
 - OS: Windows
 - Browser: Chrome, Firefox
 - Versions:
   - Pode: Pode v2.8.0
   - PowerShell: PS5.1, PS 7.2.4

### Additional Context
As per the definition of `NumericDate` in [rfc7519](https://www.rfc-editor.org/rfc/rfc7519#section-2) `exp` and `nbf` should represent "...number of seconds from 1970-01-01T00:00:00Z UTC...".

However, given the following initialization in `Cryptography.ps1`
https://github.com/Badgerati/Pode/blob/9b1456e2d03baa5a4876f551949bf1ead6be7715/src/Private/Cryptography.ps1#L264-L265

The `Kind` for these objects will be `Local` and `Unspecified` respectively. 

As per [Remarks ](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.op_equality?view=netframework-4.8#remarks )in DateTime.Equality documentation,

> The [Equality](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.op_equality?view=netframework-4.8) operator determines whether two [DateTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetime?view=netframework-4.8) values are equal by comparing their number of ticks. Before comparing [DateTime](https://learn.microsoft.com/en-us/dotnet/api/system.datetime?view=netframework-4.8) objects, make sure that the objects represent times in the same time zone. You can do this by comparing the values of their [Kind](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.kind?view=netframework-4.8) property.

The value for `Ticks` property would be different when `DateTimeKind` is specified, vs. when it is not

> PS C:\> ([datetime]::new(1970, 1, 1)).Ticks
> 621355968000000000
> PS C:\> ([datetime]::new(1970, 1, 1, 0, 0, [DateTimeKind]::Utc)).Ticks
> 621355968010000000

This makes the existing implementation of JWT validation invalid. This could be fixed by initializing `$now` as `UtcNow` and using the overload that allows specifying `DateTimeKind` for `$unixStart`. I will share a PR for this shortly in case it is helpful.

avin3sh:
### Description of the Change
Resolves #1083

### DIFF
--- a/src/Private/Cryptography.ps1
+++ b/src/Private/Cryptography.ps1
@@ -261,8 +261,8 @@ function Test-PodeJwt
         $Payload
     )
 
-    $now = [datetime]::Now
-    $unixStart = [datetime]::new(1970, 1, 1)
+    $now = [datetime]::UtcNow
+    $unixStart = [datetime]::new(1970, 1, 1, 0, 0, [DateTimeKind]::Utc)
 
     # validate expiry
     if (![string]::IsNullOrWhiteSpace($Payload.exp)) {


Badgerati (approved):
LGTM :)

PR #1117, opened 2023-07-09
Mardown style changes

ArieHein:
added new lines before code examples as mandated by the markdown styleguide :)


Badgerati:
LGTM :)

PR #1145, opened 2023-10-02
#1142 | Make the Test-PodeJwt function Public

### LINKED ISSUE #1142: Make the Test-PodeJwt function Public
### Describe the Change
The `Test-PodeJwt` function, for verifying a JWT payload's expiry dates, is current Private; it should be made Public for custom use-cases of JWTs.

alan-null:
### Description of the Change
Details in the related issue

### Related Issue
Fixes #1142

### DIFF
--- a/src/Pode.psd1
+++ b/src/Pode.psd1
@@ -218,6 +218,7 @@
         'Add-PodeAuthUserFile',
         'ConvertTo-PodeJwt',
         'ConvertFrom-PodeJwt',
+        'Test-PodeJwt'
         'Use-PodeAuth',
         'ConvertFrom-PodeOIDCDiscovery',
         'Test-PodeAuthUser',
--- a/src/Private/Cryptography.ps1
+++ b/src/Private/Cryptography.ps1
@@ -253,32 +253,6 @@ function Invoke-PodeValueUnsign
     return $raw
 }
 
-function Test-PodeJwt
-{
-    param(
-        [Parameter(Mandatory=$true)]
-        [pscustomobject]
-        $Payload
-    )
-
-    $now = [datetime]::UtcNow
-    $unixStart = [datetime]::new(1970, 1, 1, 0, 0, [DateTimeKind]::Utc)
-
-    # validate expiry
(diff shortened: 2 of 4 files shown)


Badgerati:
Hi @alan-null,

That's fine! One thing I would suggest is to move it into the public Authentication.ps1 with the other public JWT functions 😃

src/Public/Cryptography.ps1
Badgerati, line 1: As mentioned in another comment, this would be better placed in the Public/Authentication.ps1 file - it's arguably in the wrong private file to start with!

src/Public/Cryptography.ps1
```diff
@@ -0,0 +1,43 @@
...
+
+.EXAMPLE
+Test-PodeJwt @{exp = 2696258821 }
+
+.EXAMPLE
+Test-PodeJwt -Payload @{nbf = 1696258821 }
+#>
+function Test-PodeJwt
+{
+    param(
```
Badgerati, line 21: Missing a `[CmdletBinding()]` header

tests/unit/Cryptography.Tests.ps1
```diff
@@ -52,4 +52,20 @@ Describe 'New-PodeSalt' {
         Mock Get-PodeRandomBytes { return @(10, 10, 10) }
         New-PodeSalt -Length 3 | Should Be 'CgoK'
     }
+}
+
+Describe 'Test-PodeJwt' {
```
Badgerati, line 57: Along with moving the function to Authentication.ps1, this'll need relocating as well

Badgerati (changes requested):

src/Public/Cryptography.ps1
alan-null: done

src/Public/Cryptography.ps1
alan-null: done

tests/unit/Cryptography.Tests.ps1
alan-null: done

Badgerati (approved):

PR #1258, opened 2024-03-25
Fix #1257 Collection was modified after..

### LINKED ISSUE #1257: Server stops with errors "Collection was modified after..."
### Describe the Bug

Server stops with the error "Collection was modified after the enumerator was instantiated."

### Steps To Reproduce

[2024-03-25-server-stops.zip](https://github.com/Badgerati/Pode/files/14739006/2024-03-25-server-stops.zip)

(1) Start the attached `server.ps1`, rather trivial one route server.

(2) In the opened browser keep refreshing the page with brief processes details.
Alternatively, run the attached `loop.ps1` which sends similar requests in a loop.

(3) At some point, sometimes soon or after a while, the server stops with errors like:

```
Test-PodeRestartPressed: C:\Program Files\WindowsPowerShell\Modules\Pode\2.9.0\Public\Core.ps1:194
Line |
 194 |  … t.IsCancellationRequested) -or (Test-PodeRestartPressed -Key $key)) {
     |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Collection was modified after the enumerator was instantiated.
```

Errors may be at different locations and about different keys tested.

### Expected Behavior

Ideally, the error should not happen, with its reasons avoided.

If this is not possible errors should be caught and ignored.

### Platform

 - OS: Windows 11
 - Browser: Chrome
 - Versions:
   - Pode: 2.9.0
   - PowerShell: 7.4.1

mdaneri:
Server stops with errors "Collection was modified after..." #1257
```powershell
Start-PodeServer {
    Add-PodeEndpoint -Address localhost -Port 8081 -Protocol Http
    #working
    Add-PodeRoute -Path '/processes'  -Method Get -ScriptBlock {
        $r = Get-Process | .{ process { if ($_.WS -gt 100mb) { $_ } } } |
            Select-Object Name, @{e = { [int]($_.WS / 1mb) }; n = 'WS' } |
            Sort-Object WS -Descending

        Write-PodeHtmlResponse -Value $r  -StatusCode 200
    }
    

    # doesn't work
    Add-PodePage -Name NotWorkingProcesses -ScriptBlock {
        Get-Process | .{ process { if ($_.WS -gt 100mb) { $_ } } } |
            Select-Object Name, @{e = { [int]($_.WS / 1mb) }; n = 'WS' } |
            Sort-Object WS -Descending
    }
 
}
```

### Description of the Change
``` powershell
   if (Test-PodeIsEmpty $ScriptBlock) {
                throw 'A non-empty ScriptBlock is required to created a Page Route'
            }

            $arg = @($ScriptBlock, $Data)
            $logic = {
                param($script, $data)

                # invoke the function (optional splat data)
                if (Test-PodeIsEmpty $data) {
# before was $result = (. $script)
                    $result = & $script
                }
                else {
# before was $result = (. $script @data)
                    $result = & $script @data
                }

                # if we have a result, convert it to html
                if (!(Test-PodeIsEmpty $result)) {
                    Write-PodeHtmlResponse -Value $result
                }
            }
``` 
### Related Issue
#1257

### DIFF
--- a/src/Public/Routes.ps1
+++ b/src/Public/Routes.ps1
@@ -2166,10 +2166,10 @@ function Add-PodePage {
 
                 # invoke the function (optional splat data)
                 if (Test-PodeIsEmpty $data) {
-                    $result = & $script
+                    $result = Invoke-PodeScriptBlock -ScriptBlock $script -Return
                 }
                 else {
-                    $result = & $script @data
+                    $result = Invoke-PodeScriptBlock -ScriptBlock $script -Arguments $data -Return
                 }
 
                 # if we have a result, convert it to html
--- a/src/Public/Routes.ps1
+++ b/src/Public/Routes.ps1
@@ -2166,10 +2166,10 @@ function Add-PodePage {
 
                 # invoke the function (optional splat data)
                 if (Test-PodeIsEmpty $data) {
-                    $result = (. $script)
+                    $result = & $script
                 }
                 else {
-                    $result = (. $script @data)
+                    $result = & $script @data
                 }
 
                 # if we have a result, convert it to html


src/Public/Routes.ps1
```diff
@@ -2166,10 +2166,10 @@ function Add-PodePage {
 
                 # invoke the function (optional splat data)
                 if (Test-PodeIsEmpty $data) {
-                    $result = (. $script)
+                    $result = & $script
                 }
                 else {
-                    $result = (. $script @data)
+                    $result = & $script @data
                 }
```
Badgerati, line 2173: I just tested this locally myself, and even with the switch to `&` I still get the issue.

I changed the block to use `Invoke-PodeScriptBlock` instead and that seems to have fixed it on myside - and sped it up as well:
```poweshell
if (Test-PodeIsEmpty $data) {
    $result = Invoke-PodeScriptBlock -ScriptBlock $script -Return
}
else {
    $result = Invoke-PodeScriptBlock -ScriptBlock $script -Arguments $data -Return
}
```

Badgerati (changes requested):

src/Public/Routes.ps1
mdaneri: with & was working for me.
but I'm happy to change to `Invoke-PodeScriptBlock`

Badgerati (approved):
LGTM

Issue #1243, opened 2024-02-21
Enhancing Flexibility with Customizable Default Folder Paths in Pode


mdaneri:
### Introduction

Pode, currently defines three critical directories (`errors`, `views`, and `public`) that are mounted as PSDrives if they exist under the server root. These directories play pivotal roles in the application's functionality, from rendering views to serving public assets and handling errors. As it stands, the paths to these directories are predefined and immutable, which can introduce limitations for users with advanced or specific configuration needs.

### Proposal for Enhancement

We propose an enhancement to Pode that introduces the capability for users to specify custom paths for these default folders (`errors`, `views`, and `public`). This flexibility will allow users to tailor the server's directory structure to fit their deployment environments or organizational policies better.

#### Implementation Details

1. **Configuration via `server.psd1`**:
   Users can define custom paths for the default folders within the `server.psd1` configuration file. This approach maintains the configuration-driven design of Pode and ensures ease of setup. For example:

   ```powershell
   @{
       Server = @{
           DefaultFolders = @{
               Public = 'c:\custom\public'
               Views  = 'd:\shared\views'
               Errors = 'e:\logs\errors'
           }
       }
   }
   ```

2. **New PowerShell Functions**:
   To support this feature dynamically, we propose adding two new functions:
   
   - `Set-PodeDefaultFolder`: Allows users to programmatically set the path for a default folder during runtime.
     - **Parameters**:
       - `-Type`: Specifies the folder type (`Views`, `Public`, or `Errors`).
       - `-Path`: Defines the new file system path for the folder.

   - `Get-PodeDefaultFolder`: Enables querying the current path settings for the default folders.
     - **Parameters**:

#### Benefits

- **Enhanced Customization**: Users gain greater control over the server's file system structure, allowing for more flexible deployment strategies.
- **Improved Integration**: This feature facilitates Pode's integration into environments with strict directory usage policies or where directory structures are predefined.
- **Dynamic Configuration**: By enabling runtime changes to folder paths, Pode can better support dynamic environments and deployment scenarios.

### Code Changes

`Public/Core.ps1`

```powershell
<#
.SYNOPSIS
Sets the path for a specified default folder type in the Pode server context.

.DESCRIPTION
This function configures the path for one of the Pode server's default folder types: Views, Public, or Errors.
It updates the server's configuration to reflect the new path for the specified folder type.
The function first checks if the provided path exists and is a directory;
if so, it updates the `Server.DefaultFolders` dictionary with the new path.
[...]

PR #1265, opened 2024-03-25
Implement Enhancing Flexibility with Customizable Default Folder Path

mdaneri:
Implementation Details for #1243 

Configuration via server.psd1:
Users can define custom paths for the default folders within the server.psd1 configuration file. This approach maintains Pode's configuration-driven design and ensures ease of setup. For example:

```powershell
@{
    Server = @{
        DefaultFolders = @{
            Public = 'c:\custom\public'
            Views  = 'd:\shared\views'
            Errors = 'e:\logs\errors'
        }
    }
}
```
New PowerShell Functions:
To support this feature dynamically, we propose adding two new functions:

`Set-PodeDefaultFolder`: Allows users to programmatically set the path for a default folder during runtime.

Parameters:
`-Type:` Specifies the folder type (Views, Public, or Errors).
`-Path` : Defines the new file system path for the folder.

`Get-PodeDefaultFolder`: Enables querying the current path settings for the default folders.


docs/Tutorials/Routes/Utilities/StaticContent.md
```diff
@@ -6,7 +6,7 @@ Caching is supported on static content.
 
 ## Public Directory
 
-You can place static files within the `/public` directory, at the root of your server. If a request is made for a file, then Pode will automatically check the public directory first, and if found will return the back.
+You can place static files within the `/public` directory at the root of your server, which serves as the default location for static content. However, if you need to relocate this directory, you can do so programmatically using the `Set-PodeStaticFolder` function within your server script, or specify a different location in the `server.psd1` configuration file under the `Server.DefaultFolders` property. When a request is made for a file, Pode will automatically check this designated static directory first, and if the file is found, it will be returned to the requester.
```
Badgerati, line 9: I would give a very quick example, below this paragraph, of how to set a custom public folder via `Set-PodeStaticFolder` and via the server.psd1 as well - at a minimum the latter, just so there's a reference as to the structure of the DefaultFolders block in the server.psd1 config file somewhere in the docs.

Badgerati (changes requested):

docs/Tutorials/Routes/Utilities/StaticContent.md
mdaneri: sure

src/Private/Context.ps1
```diff
@@ -814,6 +821,18 @@ function Set-PodeServerConfiguration {
...
         $Context.Server.Request.BodySize = [long]$Configuration.Request.BodySize
     }
+
+    if ($Configuration.DefaultFolders) {
+        if ($Configuration.DefaultFolders.Public) {
+            $Context.Server.DefaultFolders.Public = $Configuration.DefaultFolders.Public
+        }
+        if ($Configuration.DefaultFolders.View) {
+            $Context.Server.DefaultFolders.View = $Configuration.DefaultFolders.View
+        }
```
Badgerati, line 831: Just spotted,  the 3 `.View` refs should be `.Views` :)

Badgerati (changes requested):

src/Private/Context.ps1
mdaneri: My bad.
This is the major problem with Powershell.
Nothing tells you that what you wrote is rubbish

Badgerati (approved):
LGTM

Issue #1237, opened 2024-02-03
File Browsing feature for Pode Static Route


mdaneri:
### Describe the Feature
Pode today has no option that allows browsing a directory when a default page is not available.
The idea is to add a new `-FileBrowser` parameter to `Add-PodeStaticRoute`

### Parameter
 `-FileBrowser` When supplied, If the path is a folder, instead of returning 404, will return A browsable content of the directory.

### Code Changes 

```powershell

function Write-PodeFileResponse {
    [CmdletBinding()]
    param (
        [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
        [ValidateNotNull()]
        [string]
        $Path,

        [Parameter()]
        $Data = @{},

        [Parameter()]
        [string]
        $ContentType = $null,

        [Parameter()]
        [int]
        $MaxAge = 3600,

        [Parameter()]
        [int]
        $StatusCode = 200,

        [switch]
        $Cache,

        [switch]
        $FileBrowser
    )
    # resolve for relative path
    $Path = Get-PodeRelativePath -Path $Path -JoinRoot

    # test the file path, and set status accordingly
    if (! $FileBrowser.isPresent -and !(Test-PodePath $Path -FailOnDirectory)) {
        return
    }

    # are we dealing with a dynamic file for the view engine? (ignore html)
    $mainExt = Get-PodeFileExtension -Path $Path -TrimPeriod

    # generate dynamic content
    if (![string]::IsNullOrWhiteSpace($mainExt) -and (
        ($mainExt -ieq 'pode') -or
        ($mainExt -ieq $PodeContext.Server.ViewEngine.Extension -and $PodeContext.Server.ViewEngine.IsDynamic)
        )) {
        $content = Get-PodeFileContentUsingViewEngine -Path $Path -Data $Data

        # get the sub-file extension, if empty, use original
        $subExt = Get-PodeFileExtension -Path (Get-PodeFileName -Path $Path -WithoutExtension) -TrimPeriod
        $subExt = (Protect-PodeValue -Value $subExt -Default $mainExt)

        $ContentType = (Protect-PodeValue -Value $ContentType -Default (Get-PodeContentType -Extension $subExt))
        Write-PodeTextResponse -Value $content -ContentType $ContentType -StatusCode $StatusCode
    }

    # this is a static file
    else {
        if ( Test-PodePathIsDirectory $Path) {
            # If the path is a directory and FileBrowser switch is used, generate a browsable list of files/directories
[...]

PR #1266, opened 2024-03-25
File Browsing feature for Pode Static Route

mdaneri:
Implementation of File Browser for static route + documentation

### Description of the Change 

Pode today has no option that allows browsing a directory when a default page is not available.
The idea is to add a new -FileBrowser parameter to `Add-PodeStaticRoute` and `Add-PodeStaticRouteGroup`

####Parameter 

-FileBrowser When supplied, If the path is a folder, instead of returning 404, will return A browsable content of the directory.

### Related Issue
#1237


src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+            $Path = Get-PodeRelativePath -Path $Path -JoinRoot
+            # Attempt to retrieve information about the path
+            $pathInfo = Get-Item -Path $Path -ErrorAction SilentlyContinue
+        }
+        if ($null -eq $pathInfo) {
+            # If not, set the response status to 404 Not Found
+            Set-PodeResponseStatus -Code 404
+            return
+        }
+    }
```
Badgerati, line 452: Looking at where this function gets called, I'd say that `$Path` would always be literal after Pode has already called `Get-PodeRelativePath` or `Find-PodeStaticRoute -CheckPublic` higher up the call stack. If we want to rename the parameter to `$LiteralPath` then I'm fine with that.

These lines (432-452) could then be shrunk down to the following (I've set `-ErrorAction Stop` on `Get-Item` because at this point we should be able to access a valid file, so if it fails them something on the server has gone wrong!):
```powershell
if (!(Test-PodePath $Path)) {
    return
}

$pathInfo = Get-Item -Path $Path -Force -ErrorAction Stop
```

src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+        # If directory browsing is enabled, use the directory response function
+        if ($FileBrowser.isPresent) {
+            Write-PodeDirectoryResponseInternal -Path $Path
+            return
+        }
+        else {
+            # If browsing is not enabled, return a 404 error
+            Set-PodeResponseStatus -Code 404
+            return
+        }
```
Badgerati, line 463: These could be condensed to the following (still inside the `if` statement though:
```powershell
# If directory browsing is enabled, use the directory response function, else return a 404
if ($FileBrowser.isPresent) {
    Write-PodeDirectoryResponseInternal -Path $Path
}
else {
    Set-PodeResponseStatus -Code 404
}

return
```

src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+Serves the 'logo.png' file as a static file with the specified content type and caching enabled.
+
+.OUTPUTS
+None. The function writes directly to the HTTP response stream.
+
+.NOTES
+This is an internal function and may change in future releases of Pode.
+#>
+
+function Write-PodeFileResponseInternal {
```
Badgerati, line 133: In a similar style to the comments for `Write-PodeAttachmentResponseInternal`, the Path here is usually always checked and literal higher up. I also spotted some places which could be refactored to use the new `$pathInfo` variable 😉 

The following is what I refactored the content to, and seems to work on my local:
```powershell
if (!(Test-PodePath $Path)) {
    return
}

# Attempt to retrieve information about the path
$pathInfo = Get-Item -Path $Path -force -ErrorAction Ignore

# Check if the path is a directory
if ($pathInfo.PSIsContainer) {
    # If directory browsing is enabled, use the directory response function
    if ($FileBrowser) {
        Write-PodeDirectoryResponseInternal -Path $Path
    }
    else {
        # If browsing is not enabled, return a 404 error
        Set-PodeResponseStatus -Code 404
    }

    return
}

# are we dealing with a dynamic file for the view engine? (ignore html)
# Determine if the file is dynamic and should be processed by the view engine
$mainExt = $pathInfo.Extension.TrimStart('.')

# generate dynamic content
if (![string]::IsNullOrWhiteSpace($mainExt) -and (
        ($mainExt -ieq 'pode') -or
        ($mainExt -ieq $PodeContext.Server.ViewEngine.Extension -and $PodeContext.Server.ViewEngine.IsDynamic)
    )
) {
    # Process dynamic content with the view engine
    $content = Get-PodeFileContentUsingViewEngine -Path $Path -Data $Data

    # Determine the correct content type for the response
    # get the sub-file extension, if empty, use original
    $subExt = [System.IO.Path]::GetExtension($pathInfo.BaseName).TrimStart('.')
    $subExt = (Protect-PodeValue -Value $subExt -Default $mainExt)

    $ContentType = (Protect-PodeValue -Value $ContentType -Default (Get-PodeContentType -Extension $subExt))

    # Write the processed content as the HTTP response
    Write-PodeTextResponse -Value $content -ContentType $ContentType -StatusCode $StatusCode
}
# this is a static file
else {
    if (Test-PodeIsPSCore) {
        $content = (Get-Content -Path $Path -Raw -AsByteStream)
    }
    else {
        $content = (Get-Content -Path $Path -Raw -Encoding byte)
    }

    if ($null -ne $content) {
        # Determine and set the content type for static files
        $ContentType = Protect-PodeValue -Value $ContentType -Default (Get-PodeContentType -Extension $mainExt)

        # Write the file content as the HTTP response
        Write-PodeTextResponse -Bytes $content -ContentType $ContentType -MaxAge $MaxAge -StatusCode $StatusCode -Cache:$Cache
    }
}
```

src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+            WindowsMode = $windowsMode.ToString().ToLower()
+            FileContent = $htmlContent.ToString()   # Convert the StringBuilder content to a string
+        }
+
+        $podeRoot = Get-PodeModuleMiscPath
+        # Write the response
+        Write-PodeFileResponseInternal -Path ([System.IO.Path]::Combine($podeRoot, 'default-file-browsing.html.pode')) -Data $Data
+    }
+    catch {
+        write-podehost $_
```
Badgerati, line 379: I would just let the error get bubbled up the chain, and remove the try/catch. This way the error is caught and appropriately logged and reported back to the user as a 500

src/Misc/default-file-browsing.html.pode
```diff
@@ -0,0 +1,225 @@
...
+        href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-icons/1.8.1/font/bootstrap-icons.min.css">
+
+    <title>File Browser:$($Data.Path)</title>
+    <meta charset='utf-8'>
+
+    <style>
+        h3 {
+            font-family: 'Courier New', Courier, monospace;
+            /* Applies monospace font to <h2> */
+        }
```
Badgerati, line 15: There aren't any `h3` elements on the page for this to apply, however if we replace `h3` with `body` this will match the styling with other default Pode pages, and also won't affect the monospaced table either.

The font-family will also need replacing with `Helvetica,Arial,sans-serif`.

src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+        foreach ($item in $child) {
+            $link = "$baseLink/$([uri]::EscapeDataString($item.Name))"
+            if ($item.PSIsContainer) {
+                $size = ''
+                $icon = 'bi bi-folder2'
+            }
+            else {
+                $size = '{0:N2}KB' -f ($item.Length / 1KB)
+                $icon = 'bi bi-file'
+            }
```
Badgerati, line 337: Change `$icon` to be emojis, this way we can remove the dependency on bootstrap - I also think it looks nicer as well 🤣 
```powershell
if ($item.PSIsContainer) {
    $size = ''
    $icon = '📁'
}
else {
    $size = '{0:N2}KB' -f ($item.Length / 1KB)
    $icon = '📄'
}
```

src/Private/Responses.ps1
```diff
@@ -48,4 +82,429 @@ function Show-PodeErrorPage {
...
+            $htmlContent.Append("</td> <td class='dateTime'>")
+            $htmlContent.Append($item.CreationTime.ToString('yyyy-MM-dd HH:mm:ss'))
+            $htmlContent.Append("</td> <td class='dateTime'>")
+            $htmlContent.Append($item.LastWriteTime.ToString('yyyy-MM-dd HH:mm:ss'))
+            $htmlContent.Append("</td> <td class='size'>")
+            $htmlContent.Append( $size)
+            $htmlContent.Append( "</td> <td class='icon'><i class='")
+            $htmlContent.Append( $icon)
+            $htmlContent.Append( "'></i></td> <td class='name'><a href='")
```
Badgerati, line 360: With the icon change above, this means lines 358-360 will need altering:
```powershell
$htmlContent.Append( "</td> <td class='icon'>")
$htmlContent.Append( $icon)
$htmlContent.Append( "</td> <td class='name'><a href='")
```

src/Misc/default-file-browsing.html.pode
```diff
@@ -0,0 +1,225 @@
+<html lang="en" ; style='background-color: #01b0c4' ;>
+
+<head>
+    <!-- Include the Bootstrap Icons stylesheet -->
+    <link rel="stylesheet"
+        href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-icons/1.8.1/font/bootstrap-icons.min.css">
```
Badgerati, line 6: Using emojis, this can be removed.

Badgerati (changes requested):

docs/Tutorials/Routes/Utilities/StaticContent.md
```diff
@@ -168,3 +168,40 @@ Start-PodeServer {
...
+
+```powershell
+Start-PodeServer -ScriptBlock {
+    Add-PodeEndpoint -Address localhost -Port 8080 -Protocol Http
+    Add-PodeStaticRoute -Path '/' -Source './content/assets' -FileBrowser
+    Add-PodeStaticRoute -Path '/download' -Source './content/newassets' -DownloadOnly -FileBrowser
+}
+```
+
+When used with `-Download,` the browser downloads any file selected instead of rendering. The folders are rendered and not downloaded.
```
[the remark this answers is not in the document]
mdaneri: I think this is fault of Grammarly :) there was a comma inside the ``

src/Private/Helpers.ps1
```diff
@@ -1694,35 +1710,20 @@ function Test-PodePath {
...
 
-    # if the file isn't accessible then fail 401
-    if (!(Test-PodePathAccess $Path)) {
-        if (!$NoStatus) {
-            Set-PodeResponseStatus -Code 401
-        }
-
+    # if the file doesnt exist then fail on 404
+    if ($NoStatus.IsPresent) {
         return $false
```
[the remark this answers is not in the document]
mdaneri: Test-PodePath shouldn't return any additional information other than whether it is accessible or not.
Persmission-related information is only a security concern and doesn't provide any additional helpful information to the user.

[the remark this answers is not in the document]
mdaneri: but if you think it's is a must to have I can change it.
This topic is only Windows related on Linux/MacOs you never get System.UnauthorizedAccessException

src/Private/Responses.ps1
mdaneri: I was trying to reduce the number of calls to Get-Item.
I found it redundant to use Test-PodePath. Anyway, the use of Test-PodePath will cause any  file marked as hidden to fail

src/Private/Responses.ps1
mdaneri: I was not aware that something like that will work :) 
Let me try

mdaneri: I love it

src/Private/Responses.ps1
mdaneri: It was for debug and I forgot to remove it

src/Private/Responses.ps1
mdaneri: I agree :)

src/Private/Middleware.ps1
```diff
@@ -195,14 +195,40 @@ function Get-PodePublicMiddleware {
...
+
+#>
 function Get-PodeRouteValidateMiddleware {
     return @{
         Name  = '__pode_mw_route_validation__'
         Logic = {
-            # check if the path is static route first, then check the main routes
-            $route = Find-PodeStaticRoute -Path $WebEvent.Path -EndpointName $WebEvent.Endpoint.Name
-            if ($null -eq $route) {
+            if ($Server.Configuration.Server.RouteOrderMainBeforeStatic) {
```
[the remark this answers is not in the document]
mdaneri: Do you think is an issue to reverse this policy? 
I mean having nostatic validated before static?

src/Public/Routes.ps1
```diff
@@ -749,12 +759,16 @@ function Add-PodeStaticRoute {
...
+    #The path use KleeneStar(Asterisk)
+    $KleeneStar = $OrigPath.Contains('*')
+
     # add the route(s)
     Write-Verbose "Adding Route: [$($Method)] $($Path)"
     $newRoutes = @(foreach ($_endpoint in $endpoints) {
             @{
                 Source            = $Source
                 Path              = $Path
+                KleeneStar        = $KleeneStar
```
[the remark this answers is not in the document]
mdaneri: Yes I'm not using anymore this property

src/Private/Responses.ps1
mdaneri: Same as before included the  hidden files issue

docs/Tutorials/Configuration.md
```diff
@@ -68,17 +68,18 @@ A "path" like `Server.Ssl.Protocols` looks like the below in the file:
...
+| Server.Request                    | Defines request timeout and maximum body size                           | [link](../RequestLimits)                   |
+| Server.AutoImport                 | Defines the AutoImport scoping rules for Modules, SnapIns and Functions | [link](../Scoping)                         |
+| Server.Logging                    | Defines extra configuration for Logging, like masking sensitive data    | [link](../Logging/Overview)                |
+| Server.Root                       | Overrides root path of the server                                       | [link](../Misc/ServerRoot)                 |
+| Server.Restart                    | Defines configuration for automatically restarting the server           | [link](../Restarting/Types/AutoRestarting) |
+| Server.FileMonitor                | Defines configuration for restarting the server based on file updates   | [link](../Restarting/Types/FileMonitoring) |
+| Server.RouteOrderMainBeforeStatic | Changes the way routes are processed.                                   | [link](../Routes/Utilities/StaticContent)  |
```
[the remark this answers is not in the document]
mdaneri: done

src/Private/Responses.ps1
mdaneri: Reagrding the parameter. I thought I would rename the parameter to something like Literalpath. But I changed my mind because Powershell always uses Path for everything like that

src/Private/Helpers.ps1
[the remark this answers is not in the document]
mdaneri: I was rethinking that, and maybe you are right. Let me change the function.
I need to include a -Force switch for the files with the Hidden attribute

src/Private/Responses.ps1
mdaneri: The only thing I chaged is -ErrorAction Stop and Test-Path with -Force

src/Private/Responses.ps1
mdaneri: regarding `$Path = Get-PodeRelativePath -Path $Path -JoinRoot` I remember taht I tried to remove it but someway I broke the browsing. Probably it was something else because now everything works as expected without the line

src/Private/Helpers.ps1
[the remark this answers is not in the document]
Badgerati: I just spotted a bug for the status code, totally my bad. If the Path is empty, or we FailOnDirectory, then a 200 will be set when it should be a 404! 🙈 

This should do the trick:
```powershell
$statusCode = 404
if (![string]::IsNullOrWhiteSpace($Path)) {
    try {
        $item = Get-Item $Path -Force:$Force -ErrorAction Stop
        if (!$FailOnDirectory -or !$item.PSIsContainer) {
            return $true
        }
    }
    catch [System.Management.Automation.ItemNotFoundException] {
        $statusCode = 404
    }
    catch [System.UnauthorizedAccessException] {
        $statusCode = 401
    }
    catch {
        $statusCode = 400
    }
}

# if we failed to get the file, report back the status code and/or return true/false
if (!$NoStatus.IsPresent) {
    Set-PodeResponseStatus -Code $statusCode
}

return $false
```

src/Private/Responses.ps1
Badgerati: > I was trying to reduce the number of calls to Get-Item. I found it redundant to use Test-PodePath. If I change the code the way you suggest the
> 
> ```powershell
>  if ($null -eq $pathInfo) {
>         #if not exist try with to find with public Route if exist
>         $Path = Find-PodePublicRoute -Path $Path
>         if ($Path) {
>             # only attach files from public/static-route directories when path is relative
>             $Path = Get-PodeRelativePath -Path $Path -JoinRoot
>             # Attempt to retrieve information about the path
>             $pathInfo = Get-Item -Path $Path -ErrorAction SilentlyContinue
>         }
>         if ($null -eq $pathInfo) {
>             # If not, set the response status to 404 Not Found
>             Set-PodeResponseStatus -Code 404
>             return
>         }
>     }
> ```
> 
> stop to work. this code is coming from the previous public function Write-PodeAttachmentResponse. If you tell me that is not useful I can chage the code the way you suggest

I'm looking in the original `Set-PodeResponseAttachment` but can't see this code. From what I can see with `Set-PodeResponseAttachment` now and the new `Write-PodeAttachmentResponseInternal` they roughly do the same thing, just with the `Get-Item` change - unless I'm being blind!

mdaneri: I changed the way test-podepath works.
Let me know if it's ok now

src/Private/Helpers.ps1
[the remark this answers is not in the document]
Badgerati: Happy with the `$ReturnItem` change, however the bug mentioned just above in the previous comment for status codes still exists.

docs/Tutorials/Configuration.md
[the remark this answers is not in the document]
Badgerati: In the `Configuration.md` file, `Server.RouteOrderMainBeforeStatic` was updated to be `Server.Web.Static.ValidateLast` instead of just `Web.Static.ValidateLast`.

The change to `Set-PodeWebConfuguration` is good.

docs/Tutorials/Routes/Utilities/StaticContent.md
```diff
@@ -168,3 +168,40 @@ Start-PodeServer {
...
+Nothing to report :D
+'@
+        Write-PodeTextResponse -Value $value
+    }
+
+    Add-PodeStaticRoute -Path '/' -Source "./content" -FileBrowser
+}
+```
+
+To change the default behavior, you can use the `Server.RouteOrderMainBeforeStatic` property in the `server.psd1` configuration file, setting the value to `$True.` This will ensure that any static route is evaluated after any other route.
```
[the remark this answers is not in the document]
Badgerati: Similar issue here, should be `Web.Static.ValidateLast` rather than `Server.Web.Static.ValidateLast`

src/Private/Helpers.ps1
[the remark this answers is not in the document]
mdaneri: I tried your change and anything stopped to work

src/Private/Responses.ps1
Badgerati: There are changes in the above to utilise `$pathInfo` more, rather than making calls to `Get-PodeFileExtension` for `$mainExt` or that and `Get-PodeFileName` for `subExt`. Do we want to include these, as it would make the code slightly more performant?

mdaneri: Sure why not
I'm with my phone so I cannot test anything at the moment

src/Private/Helpers.ps1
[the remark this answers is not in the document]
Badgerati: Oh? What stopped, as I have this change running locally atm and everything seems to be working?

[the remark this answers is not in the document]
mdaneri: I was sleepy due to jet lag maybe I made some mistakes 😓 let me try again when I'll be back in hotel

docs/Tutorials/Configuration.md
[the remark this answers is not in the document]
mdaneri: I see I made the wrong assumption on how Web configuration works

docs/Tutorials/Routes/Utilities/StaticContent.md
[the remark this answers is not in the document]
mdaneri: same as before

src/Private/Responses.ps1
mdaneri: this part was on the old function. But I agree with you that cannot be hit. igoing to remove it

src/Private/Helpers.ps1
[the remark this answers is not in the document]
mdaneri: should be   `if (($null -ne $item) -and (!$FailOnDirectory -or !$item.PSIsContainer)) {` not `if (!$FailOnDirectory -or !$item.PSIsContainer) {`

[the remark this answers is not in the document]
mdaneri: also there is   no  Set-PodeResponseStatus -Code 200

src/Private/Responses.ps1
```diff
@@ -48,4 +82,403 @@ function Show-PodeErrorPage {
...
+
+    # Check if the path is a directory
+    if ( $pathInfo.PSIsContainer) {
+        # If directory browsing is enabled, use the directory response function
+        Write-PodeDirectoryResponseInternal -Path $Path
+    }
+    else {
+        # are we dealing with a dynamic file for the view engine? (ignore html)
+        # Determine if the file is dynamic and should be processed by the view engine
+        $mainExt = Get-PodeFileExtension -Path $Path -TrimPeriod
```
Badgerati, line 178: This line can be replaced to use `$pathInfo` as well :)
```powershell
$mainExt = $pathInfo.Extension.TrimStart('.')
```

src/Private/Responses.ps1
```diff
@@ -48,4 +82,403 @@ function Show-PodeErrorPage {
...
+                Write-PodeTextResponse -Bytes $content -ContentType $ContentType -MaxAge $MaxAge -StatusCode $StatusCode -Cache:$Cache
+                return
+            }
+            catch [System.UnauthorizedAccessException] {
+                $statusCode = 401
+            }
+            catch {
+                $statusCode = 400
+            }
+            # If the file does not exist, set the HTTP response status to 404 Not Found
```
Badgerati, line 221: Last part of the comment needs changing, to reference something like "set the HTTP response status code appropriately"

src/Private/Helpers.ps1
```diff
@@ -1714,37 +1736,58 @@ function Test-PodePath {
...
+        }
+        else {
+            return $null
         }
-
-        return $false
     }
 
-    return $true
+    return ($statusCode -eq 200)
```
Badgerati, line 1789: The 200 status code parts at the bottom won't be hit, as the `try` returns beforehand. A 200 doesn't really need to be explicitly set, as Pode does this by default as it's default response status unless a different one is set by `Set-PodeResponseStatus`.

This could then be condensed to the following, moving the try success return further up:
```powershell
$statusCode = 404

if (![string]::IsNullOrWhiteSpace($Path)) {
    try {
        $item = Get-Item $Path -Force:$Force -ErrorAction Stop
        if (($null -ne $item) -and (!$FailOnDirectory -or !$item.PSIsContainer)) {
            if ($ReturnItem) {
                return $item
            }
            return $true
        }
    }
    catch [System.Management.Automation.ItemNotFoundException] {
        $statusCode = 404
    }
    catch [System.UnauthorizedAccessException] {
        $statusCode = 401
    }
    catch {
        $statusCode = 400
    }
}

# if we failed to get the file, report back the status code and/or return true/false
if (!$NoStatus.IsPresent) {
    Set-PodeResponseStatus -Code $statusCode
}

if ($ReturnItem) {
    return $null
}
return $false
```

Badgerati (changes requested):

src/Private/Helpers.ps1
mdaneri: because I forgot to delete it :D

Badgerati: The issue the last commit fixed, due to calling `Set-PodeResponseStatus` in `Test-PodePath`, is because you're setting a 200 response within the function, when it's not needed - the original version of this function doesn't set a 200, a it can be left to Pode's default. If you use the code snippet above everything works as expected, even the tests.

mdaneri: I understood the issue thanks. I'll fix morning

mdaneri: done

src/Private/Helpers.ps1
```diff
@@ -1766,22 +1766,21 @@ function Test-PodePath {
...
     if ($ReturnItem) {
-        if ($statusCode -eq 200) {
-            return  $item, $statusCode
-        }
-        else {
-            return  $null, $statusCode
+        # if we failed to get the file, report back the status code and/or return true/false
+        if (!$NoStatus.IsPresent) {
+            Set-PodeResponseStatus -Code $statusCode
         }
```
Badgerati, line 1780: Lines 1777-1780, for setting the status code, need to be moved up to line 1775 above the `if ($ReturnItem)` check, otherwise we're not setting the status code when needed. `$NoStatus` and `$ReturnItem` should be independent of one another.

```powershell
# if we failed to get the file, report back the status code and/or return true/false
if (!$NoStatus.IsPresent) {
    Set-PodeResponseStatus -Code $statusCode
}

if ($ReturnItem) {
    return  $null
}
return $false
```

For example, in `examples/web-pages-simple.ps1`, if you change the `Write-PodeViewResponse -Path 'simple'` to `Write-PodeViewResponse -Path 'simple2'` instead this will now return a 200 response and not a 404 like it does today.

Badgerati (changes requested):

Badgerati (approved):
LGTM 😄 I'll get it merged in a bit

PR #1262, opened 2024-03-25
Migration to Pester 5.5

mdaneri:
Migration to Pester 5.5

### Description of the Change
Test Migration to Pester 5.5

### Reference
#1260

PR #1267, opened 2024-03-25
Add DualMode support for Endpoints to listen on IPv4 and IPv6 at the same time

### LINKED ISSUE #1251: Pode cannot listen on both ipv6 and ipv4 unspecified addresses at the same time - it should be able to
### Describe the Bug
Since #915 was fixed, pode can now listen on an ipv6 unspecified address, but it cannot listen on both ipv6 and ipv4 unspecified addresses at the same time.

### Steps To Reproduce
```Powershell
Start-PodeServer -ScriptBlock {
  Add-PodeEndpoint -Address '::' -Port 8080 -Protocol http
  Add-PodeEndpoint -Address '0.0.0.0' -Port 8080 -Protocol http

  Add-PodeRoute -Method Get -Path /meat -ScriptBlock {
    Write-Host -Object (Get-Random -InputObject @('Bacon', 'Sausage', 'Steak', 'Italian Sausage', 'Beef'))
    Write-PodeTextResponse -Value success
  }
}
```

Result:
```Powershell
> Get-NetTCPConnection -State Listen -LocalPort 8080

LocalAddress                        LocalPort RemoteAddress                       RemotePort State       AppliedSetting
------------                        --------- -------------                       ---------- -----       --------------
::                                  8080      ::                                  0          Listen
```

Or, if you do ipv4 first:
```Powershell
Start-PodeServer -ScriptBlock {
  Add-PodeEndpoint -Address '0.0.0.0' -Port 8080 -Protocol http
  Add-PodeEndpoint -Address '::' -Port 8080 -Protocol http

  Add-PodeRoute -Method Get -Path /meat -ScriptBlock {
    Write-Host -Object (Get-Random -InputObject @('Bacon', 'Sausage', 'Steak', 'Italian Sausage', 'Beef'))
    Write-PodeTextResponse -Value success
  }
}
```

Result:
```Powershell
> Get-NetTCPConnection -State Listen -LocalPort 8080

LocalAddress                        LocalPort RemoteAddress                       RemotePort State       AppliedSetting
------------                        --------- -------------                       ---------- -----       --------------
0.0.0.0                             8080      0.0.0.0                             0          Listen
```

### Expected Behavior
BOTH listen addresses (IPv4 unspecified, aka 0.0.0.0, and IPv6 unspecified, aka '::') should be listening on the same port at the same time.

### Platform
 - OS: Windows Server build 20348.2322
 - Browser: N/A
 - Versions:
   - Pode: Pode 2.9.0
   - PowerShell: 5.1.20348.2227

### Additional Context
I suspect Pode is trying to dedupe the listen address before attempting to listen again.  In the interactive Pode output, both the ipv6 unspecified and ipv4 unspecified show up as:
```Powershell
Listening on the following 1 endpoint(s) [1 thread(s)]:
        - http://localhost:8080/
```

So, I would guess that Pode is resolving the address:port to `localhost:8080`, then seeing that it is already listening on `localhost:8080` and then does not attempt to create another listener.

I think this is the issue here:
https://github.com/Badgerati/Pode/blob/df86e0fb5fb37a30d684d79933b95734e4101286/src/Public/Core.ps1#L1006-L1008

But I am honestly having trouble following the logic of what hostname/friendlyname/address should/shouldn't be as that function progresses up to that point.

Badgerati:
### Description of the Change
Adds a new `-DualMode` switch on `Add-PodeEndpoint`, to enable support for listening on both IPv4 and IPv6 for an `-Address` (if IPv6 is supported).

### Related Issue
Resolves #1251 

### Examples
```powershell
Add-PodeEndpoint -Address * -Port 8080 -Protocol Http -DualMode
```


mdaneri:
You hit the issue with 7.4.1 on Windows

Badgerati:
@mdaneri What issue's that? 👀

mdaneri:
The issue is that on Windows with pwsh 7.4.x, without any apparent reason, the endpoint refuses any incoming connection.

If you go back to 7.3.x or you are using Linux or Mac, the problem never manifests itself

mdaneri:
I forgot to mention that when the test fails, it always happens with the restful test

Badgerati:
Interesting, I do have 7.4.1 locally and only tend to see it happen in a GitHub; I'll have to run it several times locally see if I can get it to trigger 🤔

mdaneri:
My Windows is virtual and fails almost every time

mdaneri:
I think I found the source of the trouble 
`Invoke-RestMethod` If I replace the cmdlet with `curl.exe`, everything works super fast without any problem.

Looks like the `Invoke-RestMethod` needs some work 
we can modify the test to use curl instead of invoke

mdaneri:
@Badgerati  I "fixed" the restful HTTPS test with 7.4.x 
the fix is committed to Migration to Pester 5.5  #1262
Please merge #1262 before your 1267, and should solve your test issue

PR #1274, opened 2024-03-31
Issue after invoke-build clean

mdaneri:
after `Invoke-Build -clean` is invoked the /src/Libs folder is removed.
the `invoke-Build -build` is not going to create it back.
this issue I think precede the `build.ps1` changes.
This should fix the issue restoring the `/src/Libs` anytime a build is invoked

### DIFF
--- a/pode.build.ps1
+++ b/pode.build.ps1
@@ -233,6 +233,10 @@ Task Build BuildDeps, {
     finally {
         Pop-Location
     }
+
+    if (Test-Path ./src/Listener/bin/Release) {
+        Copy-Item -Path ./src/Listener/bin/Release -Destination ./src/Libs -Recurse
+    }
 }
 
 


Badgerati:
I can't reproduce the same issue; after `Invoke-Build Clean` and then `Invoke-Build Build` the Libs folder is restored successfully for myself. Also, the same "clean" line is done at the start of the Build task as well 🤔 
```powershell
Task Build BuildDeps, {
    if (Test-Path ./src/Libs) {
        Remove-Item -Path ./src/Libs -Recurse -Force | Out-Null
    }
    ...
}
```

This line in Build (`Invoke-PodeBuildDotnetBuild`) is what creates the folder and then publishes the artefacts there:
```powershell
dotnet publish --configuration Release --self-contained --framework $target $AssemblyVersion --output ../Libs/$target
```

Is this line erroring for yourself?

mdaneri:
I found the issue with Pester let me do some more testing

mdaneri:
by mistake I deleted lines 126-129 during a merge 
```powershell
    dotnet publish --configuration Release --self-contained --framework $target $AssemblyVersion --output ../Libs/$target
    if (!$?) {
        throw "dotnet publish failed for $($target)"
    }
```
Because these 3 lines are like 121-123
```powershell
dotnet build --configuration Release --self-contained --framework $target $AssemblyVersion
    if (!$?) {
        throw "dotnet build failed for $($target)"
    }
```

mdaneri:
I have a question: do we need to continue to produce Listener/bin ? 
It's not used anywhere

Badgerati:
`dotnet publish` will build the `/bin` anyway, which I've also just spotted so we could actually drop the `dotnet build` part and just have `dotnet publish`.

mdaneri:
no action is required 
the drop the dotnet build part and just have dotnet publish can be part of #1276

PR #1290, opened 2024-04-22
Update Tasks.md

fatherofinvention:
Fixing a typo :)


Badgerati:
Funnily enough I had the exact same thought last week when I pushed the release notes and it ran 😂 I've got it noted down to exclude /docs, or only if /src is edited. 😄

Issue #1272, opened 2024-03-29
`Write-PodeHtmlResponse` returns the wrong result when an array of strings is piped.


mdaneri:
### Describe the Bug
`Write-PodeHtmlResponse` returns the wrong result when an array of strings is piped.
It works correctly if the array is passed as a parameter

### Steps To Reproduce
```powershell
Add-PodeRoute -Path '/processesPipedNotWorking'  -Method Get -ScriptBlock {
        Get-Process | .{ process { if ($_.WS -gt 100mb) { $_ } } } |
            Select-Object Name, @{e = { [int]($_.WS / 1mb) }; n = 'WS' } |
            Sort-Object WS -Descending | Write-PodeHtmlResponse  -StatusCode 200
    }
```

PR #1319, opened 2024-05-29
Fixes the Write-Pode(*)Response functions so the Value parameter appropriately handles when an array is passed using piping

mdaneri:
Address the #1272 bug (Write-PodeHtmlResponse returns the wrong result when an array of strings is piped.)
The same issue is present with the following functions:

- Write-PodeCsvResponse
- Write-PodeHtmlResponse
- Write-PodeJsonResponse
- Write-PodeXmlResponse
- Write-PodeYamlResponse
 

 

### Description of the Change
Added support for array values passed using piping

### Related Issue
Issue @1272

### Examples
```
Add-PodeRoute -Path '/processesPipedNotWorking'  -Method Get -ScriptBlock {
        Get-Process | .{ process { if ($_.WS -gt 100mb) { $_ } } } |
            Select-Object Name, @{e = { [int]($_.WS / 1mb) }; n = 'WS' } |
            Sort-Object WS -Descending | Write-PodeHtmlResponse  -StatusCode 200
    }
```

### DIFF
--- a/src/Public/Responses.ps1
+++ b/src/Public/Responses.ps1
@@ -474,7 +474,7 @@ function Write-PodeCsvResponse {
             }
 
             'value' {
-                if ($pipelineValue -and $pipelineValue.Count -gt 1) {
+                if ($pipelineValue.Count -gt 1) {
                     $Value = $pipelineValue
                 }
 
@@ -560,7 +560,7 @@ function Write-PodeHtmlResponse {
             }
 
             'value' {
-                if ($pipelineValue -and $pipelineValue.Count -gt 1) {
+                if ($pipelineValue.Count -gt 1) {
                     $Value = $pipelineValue
                 }
                 if ($Value -isnot [string]) {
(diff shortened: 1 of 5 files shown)


src/Public/Responses.ps1
```diff
@@ -830,30 +908,48 @@ function Write-PodeYamlResponse {
...
+            'value' {
+                if ($pipelineValue) {
+                    $Value = $pipelineValue
+                }
+
+                if ($Value -is [hashtable]) {
+                    $Value = @(foreach ($v in $Value) {
+                            New-Object psobject -Property $v
+                        })
+                }
```
Badgerati, line 939: Is this needed for YAML? Because `ConvertTo-PodeYamlInternal` handles hashtables 🤔

Badgerati (changes requested):

src/Public/Responses.ps1
mdaneri: no yaml doesn't need it like json.
only CSV and XML need it

mdaneri: this was a refuse from my tests.
it's useless and it's not required

Badgerati (approved):
LGTM

Issue #1329, opened 2024-06-07
PodeRouteGroup with a file path


eltyBelgium:
### Question
Is the following possible on any way? 

Server.ps1
```powershell
 Add-PodeRouteGroup -Path '/network' -FilePath './Routes/File.ps1'
```

File.ps1
```powershell
{
    Add-PodeRoute -Method Get -Path '/domains' -ScriptBlock {
        $result = vsz-getdomains | ConvertTo-Json

        Write-PodeJsonResponse -Value $result
    }

    Add-PodeRoute -Method Get -Path '/zones' -ScriptBlock {
        $result = vsz-getzones | ConvertTo-Json

        Write-PodeJsonResponse -Value $result
    }

    Add-PodeRoute -Method Get -Path '/zones/:zoneId/wlans' -ScriptBlock {
        $result = vsz-getwlans $WebEvent.Parameters['zoneId']| ConvertTo-Json

        $specificWlan = $result | Where-Object { $_.ZoneID -eq $WebEvent.Query['wlandId'] }

        if($null -eq $specificWlan) {
            Write-PodeJsonResponse -Value $result
            
        }else{
            Write-PodeJsonResponse -Value $specificWlan
        }
    }

    Add-PodeRoute -Method Get -Path '/zones/:zoneId/dpsks' -ScriptBlock {
        $result = vsz-getdpsks $WebEvent.Parameters['zoneId']| ConvertTo-Json

        $specificWlan = $result | Where-Object { $_.ZoneID -eq $WebEvent.Query['dpsk'] }

        if($null -eq $specificWlan) {
            Write-PodeJsonResponse -Value $result
            
        }else{
            Write-PodeJsonResponse -Value $specificWlan
        }
    }

    Add-PodeRoute -Method Post -Path '/zones/:zoneId/dpsks' -ScriptBlock {

        try{
            $result = vsz-createdpsks -zoneID $WebEvent.Parameters['zoneId'] -wlanID $WebEvent.Data.wlandID -username $WebEvent.Data.username -userrole $WebEvent.Data.userrole
            Write-PodeTextResponse -Value $result -ContentType "application/json"
        }
        catch{
            Write-PodeErrorLog -Exception $_.Exception
            Write-PodeJsonResponse -Value @{"message" = $_.Exception.Message} -StatusCode 500
        }

    }
}
```

PR #1331, opened 2024-06-07
Issue #1329

eltyBelgium:
### Description of the Change
Add Filepath parameter to the command PodeRouteGroup

### Related Issue
[A link/reference to the GitHub issue.](https://github.com/Badgerati/Pode/issues/1329)

### DIFF
--- a/src/Public/Routes.ps1
+++ b/src/Public/Routes.ps1
@@ -1164,11 +1164,22 @@ function Add-PodeRouteGroup {
         $AllowAnon,
 
         [string[]]
-        $OADefinitionTag
+        $OADefinitionTag,
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'file')]
+        [string]
+        $FilePath
     )
 
+     
+
     if (Test-PodeIsEmpty $Routes) {
+                if ($PSCmdlet.ParameterSetName -ieq 'file') {
+                    $Routes = Convert-PodeFileToScriptBlock -FilePath $FilePath
+                }else{
+                
         throw 'No scriptblock for -Routes passed'
+                }
     }
 
     if ($Path -eq '/') {
--- a/src/Public/Routes.ps1
+++ b/src/Public/Routes.ps1
@@ -1095,8 +1095,13 @@ An Array of strings representing the unique tag for the API specification.
 This tag helps in distinguishing between different versions or types of API specifications within the application.
 You can use this tag to reference the specific API documentation, schema, or version that your function interacts with.
 
+.PARAMETER FilePath
+A literal, or relative, path to a file containing a ScriptBlock for the Route's main logic.
+
 .EXAMPLE
 Add-PodeRouteGroup -Path '/api' -Routes { Add-PodeRoute -Path '/route1' -Etc }
+.EXAMPLE
+Add-PodeRouteGroup -Path '/api' -FilePath '/routes/file.ps1'
 #>
 function Add-PodeRouteGroup {
     [CmdletBinding()]
@@ -1171,15 +1176,16 @@ function Add-PodeRouteGroup {
         $FilePath
     )
 
-     
+
 
     if (Test-PodeIsEmpty $Routes) {
-                if ($PSCmdlet.ParameterSetName -ieq 'file') {
-                    $Routes = Convert-PodeFileToScriptBlock -FilePath $FilePath
-                }else{
-                
-        throw 'No scriptblock for -Routes passed'
-                }
+        if ($PSCmdlet.ParameterSetName -ieq 'file') {
+            $Routes = Convert-PodeFileToScriptBlock -FilePath $FilePath
+        }
+        else {
+
+            throw 'No scriptblock for -Routes passed'
+        }
     }
 
     if ($Path -eq '/') {


mdaneri:
You should do the same with Add-PodeSignalRouteGroup.

eltyBelgium:
I will update it then... don't merge it yet :D

eltyBelgium:
> You should do the same with Add-PodeSignalRouteGroup.

This one is added now! 😄

src/Public/Routes.ps1
```diff
@@ -1588,11 +1612,22 @@ function Add-PodeSignalRouteGroup {
...
         [string]
-        $IfExists = 'Default'
+        $IfExists = 'Default',
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'file')]
+        [string]
+        $FilePath
     )
 
+    i
```
Badgerati, line 1622: Erroneous `i` needs removing :)

src/Public/Routes.ps1
```diff
@@ -1164,11 +1170,23 @@ function Add-PodeRouteGroup {
         $AllowAnon,
 
         [string[]]
-        $OADefinitionTag
+        $OADefinitionTag,
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'file')]
```
Badgerati, line 1175: The `[Parameter()]` block for the `$Routes` parameter for `Add-PodeRouteGroup` will need a `ParameterSetName = 'Script'` property being set as well.

And then at the top of the function, the `[CmdletBinding()]` block will need a `DefaultParameterSetName = 'Script'` property being set.

This way, we allow only `Routes` or `FilePath` to be supplied and not both.

src/Public/Routes.ps1
```diff
@@ -1588,11 +1612,22 @@ function Add-PodeSignalRouteGroup {
         [Parameter()]
         [ValidateSet('Default', 'Error', 'Overwrite', 'Skip')]
         [string]
-        $IfExists = 'Default'
+        $IfExists = 'Default',
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'file')]
```
Badgerati, line 1617: The `[Parameter()]` block for the `$Routes` parameter for `Add-PodeSignalRouteGroup` will need a `ParameterSetName = 'Script'` property being set as well.

And then at the top of the function, the `[CmdletBinding()]` block will need a `DefaultParameterSetName = 'Script'` property being set.

This way, we allow only `Routes` or `FilePath` to be supplied and not both.

Badgerati (changes requested):
Looking good :)

Akin to the `Add-PodeSignalRouteGroup` above, the `Add-PodeStaticRouteGroup` will also need the same too.

It'll also be worth adding a quick mention to the new parameter in the docs at `/docs/Tutorials/Routes/Utilities/RouteGrouping.md`. At the bottom of the `## Routes` section, a quick nod the the `-FilePath` parameter and an example will do 😄

src/Public/Routes.ps1
mdaneri: @Badgerati, why does `Add-PodeStaticRouteGroup` need the parameter? 
Static routes have no script-blocks associated with them

Badgerati: `Add-PodeStaticRoute` doesn't, but `Add-PodeStaticRouteGroup` does as it has the `-Routes` scriptblock which is invoked for adding child static routes - similar to `Add-PodeRouteGroup` and `Add-PodeSignalRouteGroup`

src/Public/Routes.ps1
eltyBelgium: Fixed in next commit

src/Public/Routes.ps1
eltyBelgium: added

src/Public/Routes.ps1
eltyBelgium: added

src/Public/Routes.ps1
```diff
@@ -1433,12 +1453,21 @@ function Add-PodeStaticRouteGroup {
         $DownloadOnly,
 
         [switch]
-        $RedirectToDefault
+        $RedirectToDefault,
+
+        [Parameter(Mandatory = $true, ParameterSetName = 'File')]
+        [string]
+        $FilePath
```
Badgerati, line 1460: This `-FilePath` parameter is missing the `.PARAMETER` definition above the function

src/Public/Routes.ps1
Badgerati, line 1: For the new `-FilePath` parameter, it'd be worth adding a quick mention of the parameter in the docs found at `/docs/Tutorials/Routes/Utilities/RouteGrouping.md`. At the bottom of the `## Routes` section, a quick nod the the `-FilePath` parameter and an example will do 😄

Badgerati (changes requested):
Just some documentation to be added 😄

eltyBelgium:
> Just some documentation to be added 😄

I will fix them, sorry for the late response, I don't get notifications from Github... checking what is going on there .

src/Public/Routes.ps1
Badgerati: Hi @eltyBelgium, the examples were added, but the `.PARAMETER` is missing - it's why the builds are all failing :)

mdaneri:
`Add-PodeStaticRouteGroup` function is missing the SYNOPSIS description for FilePath parameter
```
.PARAMETER FilePath
A literal, or relative, path to a file containing a ScriptBlock for the Route's main logic.
```

Issue #1018, opened 2022-10-06
Add Native Logging to Splunk 


jbaechtelMT:
It would be great to have a logging option to write logs directly to Splunk

Badgerati:
Hi @jbaechtelMT,

Splunk I haven't considered before, but Azure Log Analytics and AWS Cloudwatch I've though about. I'll have to go through the Splunk docs some time :)

PR #1335, opened 2024-06-14
Add New Logging Features: Syslog and Restful Support

mdaneri:
## Description:
This PR introduces new features to the Pode logging system, adding support for Syslog and Restful logging methods. The following new functions have been added:

- `Enable-PodeMainLogging`
- `Enable-PodeGeneralLogging`
- `Disable-PodeMainLogging`
- `Disable-PodeGeneralLogging`

Additionally, 

the New-PodeLoggingMethod function now supports two new type:  Syslog and Restful.
Write-PodeLog now has been improved 

## New Features:

- Logging Runspace: Logging now operates in its own runspace for improved performance and isolation.
- Service Shutdown on Logging Failure: Logging failure can be configured to trigger a service shutdown.
- Configurable Max Logging Queue Size: A new configurable maximum logging queue size has been introduced, which can trigger an error if the queue size is exceeded.
- Multiple Logging Output Methods: It is now possible to assign multiple logging output methods to a specific logging configuration.
- Multiple Log Formats: The Enable-PodeRequestLogging function now supports multiple log formats: 'Extended', 'Common', 'Combined', 'JSON'.
- Configurable Log Date Format: The log date format is now configurable.
- Syslog Transport Methods: Syslog support includes TCP, UDP, and TLS transport methods.
- Switched the internal core structure from using an ArrayList with manual locking to the threadsafe System.Collections.Concurrent.ConcurrentQueue

Here is an example of how to configure multiple logging output methods:

```powershell
$logging = @(
    (New-PodeLoggingMethod -Terminal),
    (New-PodeLoggingMethod -Syslog -Server 127.0.0.1 -Transport UDP -AsUTC -ISO8601 -FailureAction Report),
    (New-PodeLoggingMethod -File -Name 'requests' -MaxDays 4)
)

$logging | Enable-PodeRequestLogging

```


Badgerati:
Looking at `Get-PodeLoggingSysLogMethod`, I'd say that Splunk and LogInsight would better being their own separate log methods - they share some common parameters with syslog (server, port, etc.), but they don't use/send the syslog format.

Ie, On `New-PodeLoggingMethod`, new switches for:
* `-Splunk` - using the Server, Port, Token, and Hostname params
* `-LogInsight` - using the Server, Port, and Id params

And then move them out of `Get-PodeLoggingSysLogMethod`, and into Splunk/LogInsight specific functions. This would open the doors for Azure Log Analytics, AWS Cloudwatch, DataDog, and others too down the line.

For `-FailureAction`, I actually like this param, and I'm wondering if instead of being syslog specific we allow it for all method types 🤔. This means the `HandleFailure` would be best as its own function rather than a child function.

---

In terms of Splunk/LogInsight using the REST API, I've no issues - in fact I was going to do this for Azure Log Analytics. However, the methods will need to handle batch logging, more so if the REST API supports sending multiple logs in one request, rather than a single log at a time. There's an example of batch logs in `Get-PodeLoggingEventViewerMethod`, where the `$item` can either be a single item or an array - if you convert it to an array each time you can simply iterate over all items, or send them in bulk.

Badgerati:
Splunk was also raised a while back, and one recommendation was also the REST API; in fact, the example there an yours are basically the same 😄 

Here: https://github.com/Badgerati/Pode/issues/1018

src/Public/Logging.ps1
```diff
@@ -142,7 +179,51 @@ function New-PodeLoggingMethod {
...
+        [ValidateSet('UDP', 'TCP', 'TCPwithTLS', 'Splunk', 'LogInsight')]
+        $Transport = 'UDP',
+
+        [Parameter( ParameterSetName = 'Syslog')]
+        [System.Security.Authentication.SslProtocols]
+        $TlsProtocol = [System.Security.Authentication.SslProtocols]::Tls12,
+
+        [Parameter( ParameterSetName = 'Syslog')]
+        [switch]
+        $RFC3164,
```
Badgerati, line 206: This might be better as a `SyslogFormat` param maybe, with a ValidateSet of RFC3184 and RFC5424 - with a default of RFC5424. Mostly in case a newer format is released

src/Public/Logging.ps1
```diff
@@ -142,7 +179,51 @@ function New-PodeLoggingMethod {
...
+        [Parameter(Mandatory = $true, ParameterSetName = 'Syslog')]
+        [string]
+        $Server,
+
+        [Parameter( ParameterSetName = 'Syslog')]
+        [Int16]
+        $Port = 514,
+
+        [Parameter( ParameterSetName = 'Syslog')]
+        [ValidateSet('UDP', 'TCP', 'TCPwithTLS', 'Splunk', 'LogInsight')]
```
Badgerati, line 197: `TCPwithTLS` seems to be just `TLS` in most places, I'm guessing this is supposed to be just `TLS` a well?

src/Private/Logging.ps1
```diff
@@ -75,6 +75,220 @@ function Get-PodeLoggingFileMethod {
...
+        }
+
+        # Mask values
+        $item = ($item | Protect-PodeLogItem)
+        if (('UDP' , 'TCP' , 'TLS') -contains $options.Transport) {
+            $processId = $PID
+
+            # Define the facility and severity
+            $facility = 1 # User-level messages
+            $severity = 6 # Informational
```
Badgerati, line 144: On the `param($item, $options)` for this scriptblock, as can be seen in `Get-PodeLoggingEventViewerMethod`, if you add a `$rawItem` at the end you'll have access to the `.Level` of the log item. Something you might be able to parse/convert to a more relevant severity code :)

src/Public/Logging.ps1
mdaneri: yes I changed everywhere but I forgot here.
this is still a work in progress :)

Badgerati:
I've done a quick scan of the code, and can't immediately see anything concerning with the code changes. I've a couple questions, but I'll see if I answer them myself when a review properly.

Only one thing I did spot was the removal of the `dd/MMM/yyyy:HH:mm:ss zzz` format. I'm happy with the DateFormat parameter, but this date formatting needs adding back in for the Request logging as it's the format expected by the Combined Log Format - and would also make it a breaking change :)

mdaneri:
For compatibility I can set that as default format

mdaneri:
i 'm going to add additional options to support common log format and extended log format for the access log

mdaneri:
Almost done.
what is missing is the remove of a log and some doc update

PR #1340, opened 2024-06-24
Add configuration parameter `Web.OpenApi.UsePodeYamlInternal`

mdaneri:
## Description

This pull request adds a new configuration parameter `Web.OpenApi.UsePodeYamlInternal` to the Pode module. This parameter forces the use of the internal YAML converter even if `PSYaml` or `powershell-yaml` modules are available. Additionally, it optimizes the YAML conversion process by caching the module import status.

## Changes

- Added the `Web.OpenApi.UsePodeYamlInternal` configuration parameter.
- Updated the `ConvertTo-PodeYaml` function to utilize this new parameter.

## New Code

```powershell
<#
.SYNOPSIS
    creates a YAML description of the data in the object - based on https://github.com/Phil-Factor/PSYaml

.DESCRIPTION
    This produces YAML from any object you pass to it.

.PARAMETER Object
    The object that you want scripted out. This parameter accepts input via the pipeline.

.PARAMETER Depth
    The depth that you want your object scripted to

.EXAMPLE
    Get-PodeOpenApiDefinition|ConvertTo-PodeYaml
#>
function ConvertTo-PodeYaml {
    [CmdletBinding()]
    [OutputType([string])]
    param (
        [parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $true)]
        [AllowNull()]
        $InputObject,

        [parameter()]
        [int]
        $Depth = 16
    )

    begin {
        $pipelineObject = @()
    }

    process {
        $pipelineObject += $_
    }

    end {
        if ($pipelineObject.Count -gt 1) {
            $InputObject = $pipelineObject
        }

        if ($PodeContext.Server.Web.OpenApi.UsePodeYamlInternal) {
            return ConvertTo-PodeYamlInternal -InputObject $InputObject -Depth $Depth -NoNewLine
        }

        if ($null -eq $PodeContext.Server.InternalCache.YamlModuleImported) {
            $PodeContext.Server.InternalCache.YamlModuleImported = ((Test-PodeModuleInstalled -Name 'PSYaml') -or (Test-PodeModuleInstalled -Name 'powershell-yaml'))
        }

        if ($PodeContext.Server.InternalCache.YamlModuleImported) {
            return ($InputObject | ConvertTo-Yaml)
        }
        else {
            return ConvertTo-PodeYamlInternal -InputObject $InputObject -Depth $Depth -NoNewLine
        }
    }
}
```

## Reason for Change

The addition of this configuration parameter allows users to force the use of the internal YAML converter, which can be necessary for portability or compatibility reasons.


src/Private/Context.ps1
```diff
@@ -938,6 +938,13 @@ function Set-PodeWebConfiguration {
         }
     }
 
+    if ($Configuration.OpenApi -and $Configuration.OpenApi.ContainsKey('UsePodeYamlInternal')) {
+        if (!$Context.Server.Web.ContainsKey('OpenApi')) {
+            $Context.Server.Web.OpenApi = @{}
+        }
+        $Context.Server.Web.OpenApi.UsePodeYamlInternal = $Configuration.OpenApi.UsePodeYamlInternal
+    }
```
Badgerati, line 946: Any reason to not move this up to the main web config hashtable definition above, as the following, just below the `Compression = @{`? It should equate to the same, and default to `$false` as well

```powershell
OpenApi = @{
    UsePodeYamlInternal = [bool]$Configuration.OpenApi.UsePodeYamlInternal
}
```

Badgerati (changes requested):

src/Private/Context.ps1
mdaneri: because the  $Context.Server.Web.OpenApi, in theory, can be an existing key

Badgerati: `$Context.Server.Web` gets created for the first time in the block just above this one, at line 910, so the `OpenApi` key wouldn't exist at this point

mdaneri: Let me check probably you are right

Badgerati (approved):

PR #1352, opened 2024-07-06
Update MIME types to comply with RFC9512 and RFC5323

mdaneri:
**Details:**

- Updated MIME type handling to accommodate the new `application/yaml` type as specified in RFC9512.
- Changed `application/x-yaml` and `text/x-yaml` to `application/yaml`.
- Changed `text/xml` to `application/xml` in accordance with the recommendations of RFC5323.

These changes ensure that Pode is compliant with the latest MIME type standards and improves compatibility with applications that rely on these MIME types.

**Changes:**

- Added support for `application/yaml`.
- Replaced all instances of `application/x-yaml` and `text/x-yaml` with `application/yaml`.
- Replaced all instances of `text/xml` with `application/xml`.


src/Public/Responses.ps1
```diff
@@ -863,7 +883,8 @@ A String, PSObject, or HashTable value. For non-string values, they will be conv
 The path to a YAML file.
 
 .PARAMETER ContentType
-Because JSON content has not yet an official content type. one custom can be specified here (Default: 'application/x-yaml' )
+Because JSON content has not yet an official content type. one custom can be specified here (Default: 'application/yaml' )
```
Badgerati, line 886: Should be "Because YAML content" etc.

src/Public/Responses.ps1
```diff
@@ -755,6 +765,10 @@ A String, PSObject, or HashTable value.
 .PARAMETER Path
 The path to an XML file.
 
+.PARAMETER ContentType
+Because JSON content has not yet an official content type. one custom can be specified here (Default: 'application/xml' )
```
Badgerati, line 769: Should be "Because XML content" etc.

Badgerati (changes requested):
Besides the 2 minor change comments, everything else LGTM

src/Public/Responses.ps1
mdaneri: opps copy and paste virus hit again :D

Badgerati (approved):

PR #1356, opened 2024-07-11
Dutch language support

mdaneri:
### Description of the Change
Added Dutch language 
Fix the description for the language test


src/Locales/nl/Pode.psd1
```diff
@@ -0,0 +1,286 @@
+@{
```
Badgerati, line 1: A lot of this file looks to be in English, with only some translations in Dutch, is that right?

Badgerati (changes requested):

src/Locales/nl/Pode.psd1
mdaneri: Yes you are right let me update this list

mdaneri: I don't understand the reason, but ChatGPT was reluctant to translate the entire file

Badgerati (approved):

PR #1373, opened 2024-08-01
Dynamically load the Pode DLL relevant to the version of PowerShell

mdaneri:
### Description of the Change
Small quality of life improvements

### Pode.build.ps1 Test
The build task is invoked only if no DLL is present

### Pode.psm1
Not anymore, a long list of `if ($PSversion -eq xx) {Add-Type -LiteralPath yy.dll}` . Now automatically search for the best DLL version fit for the running .NET  framework


mdaneri:
This is what I found searching for this topic:

When using a DLL compiled in .NET Standard 2.0 within a .NET 8 application, there are some considerations regarding performance and compatibility:

1. **Compatibility Layer**: .NET Standard 2.0 is designed to be a compatibility layer, which means it supports a broad range of platforms, including older .NET Framework versions, .NET Core, and newer .NET versions. However, it does not take full advantage of the performance optimizations and features available in .NET 8. When running in a .NET 8 environment, your application might not be able to utilize the latest improvements provided by .NET 8, potentially affecting performance.

2. **JIT Compilation**: At runtime, .NET 8 will use Just-In-Time (JIT) compilation to optimize the execution of your code. The performance differences are generally minor unless your code heavily relies on newer features and optimizations specific to .NET 8. The older .NET Standard 2.0 code may not benefit from these JIT optimizations fully.

3. **API Performance Improvements**: .NET 8 contains performance enhancements and new APIs that are not available in .NET Standard 2.0. If your .NET Standard DLL uses older APIs or patterns, it may miss out on these optimizations. Compiling directly to .NET 8 allows your code to use these newer, faster APIs.

4. **Memory Management**: .NET 8 has improvements in memory management, garbage collection, and overall runtime performance. A DLL compiled in .NET Standard 2.0 may not directly benefit from these improvements as much as a DLL compiled in .NET 8.

5. **Startup Time and Trimming**: .NET 8 provides advanced options for trimming unused code and optimizing startup times, especially for AOT (Ahead-of-Time) compiled applications. These optimizations may not apply fully to .NET Standard DLLs.

**Conclusion**: From a performance perspective, recompiling your DLL to .NET 8 will likely yield some improvements, especially if your code can leverage newer APIs and runtime enhancements. The performance difference may not always be significant but could be noticeable in performance-critical scenarios. If maximum performance and the latest features are important, consider recompiling the DLL to .NET 8.

src/Pode.psm1
```diff
@@ -90,15 +90,25 @@ try {
...
+                    break
+                }
+            }
         }
-        else {
+
+        if (-not $loaded) {
             Add-Type -LiteralPath "$($root)/Libs/netstandard2.0/Pode.dll" -ErrorAction Stop
         }
+
```
Badgerati, line 111: The below should do the same, and avoid Interop :)

```powershell
# fetch the .net version and the libs path
$version = [System.Environment]::Version.Major
$libsPath = "$($root)/Libs"

# filter .net dll folders based on version above, and get path for latest version found
if (![string]::IsNullOrWhiteSpace($version)) {
    $netFolder = Get-ChildItem -Path $libsPath -Directory -Force |
        Where-Object { $_.Name -imatch "net[1-$($version)]" } |
        Sort-Object -Property Name -Descending |
        Select-Object -First 1 -ExpandProperty FullName
}

# use netstandard if no folder found
if ([string]::IsNullOrWhiteSpace($netFolder)) {
    $netFolder = "$($libsPath)/netstandard2.0"
}

# append Pode.dll and mount
Add-Type -LiteralPath "$($netFolder)/Pode.dll" -ErrorAction Stop
```

pode.build.ps1
```diff
@@ -501,15 +501,51 @@ Task TestNoBuild TestDeps, {
...
         Write-Output "Restore UICulture to $originalUICulture"
         # restore original UICulture
         [System.Threading.Thread]::CurrentThread.CurrentUICulture = $originalUICulture
     }
 }, PushCodeCoverage, CheckFailedTests
 
-# Synopsis: Run tests after a build
-Task Test Build, TestNoBuild
+# Synopsis: Run tests after a build if needed
+Task Test {
```
Badgerati, line 512: But there's already a `TestNoBuild` to do this, and skip the building if not needed? `Test` should always build, otherwise it won't be obvious this isn't occurring to people; if they run `Test`, change the .NET code, and run `Test` again, it wouldn't rebuild and use the new code.

If people want to run the tests without building, `TestNoBuild` is available.

Badgerati (changes requested):

pode.build.ps1
mdaneri: I forgot:(

src/Pode.psm1
mdaneri: your code is better I agree.

mdaneri:
At the end the code is yours. I haven't done anything here :D

Badgerati (approved):

PR #1382, opened 2024-09-07
Convert from pode yaml

mdaneri:
### Description
This function is doing exactly the reverse of ConvertTo-PodeYaml.
But the logic is the same if an external module is available automatically use it 

Also ConvertTo-PodeYaml function is now implemented in C# - Powershell 5.x still uses the powershell implementation


Badgerati:
My issue with moving PowerShell code directly into .NET, and using the PS.SDK library, is that there are many edge cases where things could break. For example, if anyone is stuck using PS6, 7, or 7.1, this will break them. Then there's the situation of somebody using PS7.4.4 and we bump the PS.SDK to 7.4.5, does this break them as well?

This is why I've always kept anything that's PowerShell specific in PowerShell, and then anything that is .NET standard, if it needs to be for performance reasons or needs a feature PowerShell fundamentally doesn't support, moved into the .NET DLL.

Is there a reason you've moved the YAML conversion into .NET? When testing originally its performance seemed alright to me, and I ran it against some large objects to convert.
If parts of it do suffer, can we extract just the parts that suffer into .NET, instead of everything?

mdaneri:
I get it, it's frustrating. That’s why I haven’t been pushing hard on this pull request. If you check the code, you'll notice that when netstandard2.0 is the .NET version, the YAML converter isn’t compiled, so Pode falls back to using the "old" PowerShell version.

PR #1364, opened 2024-07-23
Naming Runspaces

mdaneri:
### Description of the Change
Assign a name to each runspace

### Related Issue
#1363

Issue #1363, opened 2024-07-22
Naming Runspaces


HeyItsGilbert:
### Describe the Change
When inspecting a Pode server we often see several runspaces in different states. I know there are several runspaces for schedules and route listenings. It would be useful if the runspaces where named.

#### Currently
1. Find the PODE Process ID
2. `Enter-PSHostProcess -Id <pid>`
3.  `Get-Runspace`

```powershell
[Process:9048]: PS > Get-Runspace

 Id Name            ComputerName    Type          State         Availability
 -- ----            ------------    ----          -----         ------------
  1 Runspace1       localhost       Local         Opened        Available
  2 Runspace2       localhost       Local         Opened        Busy
  4 Runspace4       localhost       Local         Opened        Busy
  5 Runspace5       localhost       Local         Opened        Busy
  6 Runspace6       localhost       Local         Opened        Busy
  8 Runspace8       localhost       Local         Opened        Busy
 10 Runspace10      localhost       Local         Opened        Busy
 11 Runspace11      localhost       Local         Opened        Busy
 12 Runspace12      localhost       Local         Opened        Available
 13 Runspace13      localhost       Local         Opened        Busy
 15 Runspace15      localhost       Local         Opened        Busy
 16 Runspace16      localhost       Local         Opened        Busy
 17 Runspace17      localhost       Local         Opened        Busy
 18 Runspace18      localhost       Local         Opened        Busy
 19 RemoteHost      localhost       Local         Opened        Busy
 25 Runspace25      localhost       Local         Opened        Available
 32 Runspace32      localhost       Local         Opened        Available
 37 Runspace37      localhost       Local         Opened        Available
```

If we had runspaces that were labeled for schedule tasks that may help troubleshooting, etc.

HeyItsGilbert:
You beat me to it @mdaneri! lol

mdaneri:
I think it's done.
The only one that are not  named are the user tasks and schedule

HeyItsGilbert:
I was actually hoping those would be part of them, but the helper should suffice.

mdaneri:
you can assign a name.
I wrote this on the documentation


## Customizing and Managing Runspace Names

### Distinguishing Runspace Names in Pode

[...]

PR #1384, opened 2024-09-07
Customizing and Managing Runspace Names

mdaneri:
Recovered from Naming Runspaces #1364

### Description of the Change
Assign a name to each runspace

### Related Issue
#1363 

### Distinguishing Runspace Names in Pode

In Pode, internal runspaces are automatically assigned distinct names. This naming convention helps in identifying and managing these runspaces efficiently during debugging and monitoring. However, this is not the case for runspaces created by user tasks and schedules (excluding AsyncTask). These user-created runspaces typically have names in the `Runspace<number>` format, which can make it challenging to distinguish between different runspaces.

### Customizing Runspace Names

By default, Pode’s Tasks, Schedules, and Timers use their own names to label their associated runspaces, making it easier to identify them during debugging and monitoring.

However, if you wish to use a different name for the runspace, you can invoke `Add-PodeTask`, `Add-PodeSchedule`, or `Add-PodeTimer` with the `-DisableRunspaceNaming` parameter.
Then, within the script block, you can use the Set-PodeCurrentRunspaceName cmdlet to assign any custom name you prefer.

Another useful cmdlet is `Get-PodeCurrentRunspaceName`, which retrieves the current runspace's name. This can be helpful if you need to log or display the runspace name dynamically.

### Example

Here is an updated example demonstrating how to set a custom runspace name in a Pode task:

```powershell
Add-PodeTask -Name 'Test2' -DisableRunspaceNaming -ScriptBlock {
    param($value)
    # Set a custom name for the current runspace
    Set-PodeCurrentRunspaceName -Name 'My Fancy Runspace'
    Start-Sleep -Seconds 10
    "A $($value) is never late, it arrives exactly when it means to" | Out-Default
}
```

In this example, the `Set-PodeCurrentRunspaceName` cmdlet is used to assign the custom name 'My Fancy Runspace' to the runspace executing the task, enhancing identification during debugging or in logs.

### Retrieving Runspace Names

Another useful cmdlet is `Get-PodeCurrentRunspaceName`, which retrieves the current runspace's name. This cmdlet can be particularly helpful if you need to log or display the runspace name dynamically during execution. It allows you to track and manage runspaces effectively, especially in complex scenarios where multiple runspaces are running concurrently.

```powershell
 Add-PodeSchedule -Name 'test' -Cron '@hourly' -ScriptBlock { Write-PodeHost "Runspace name: $(Get-PodeCurrentRunspaceName)" }
 ```

### CODE REVIEW

**Comment by Badgerati** [comment]
I like the idea of `Set-PodeCurrentRunspaceName` and `Get-PodeCurrentRunspaceName`, but I think the initial setting of all Runspace names should be done at the point of `Add-PodeRunspace`. We can still re-use `Set-PodeCurrentRunspaceName`, but this will allow use to centralise the setting of the initial Runspace names here, and will let us also standardise on an internal naming structure as well - saves use having to call `Set-PodeCurrentRunspaceName` everywhere internally when `Add-PodeRunspace` can do it for us. (or more precisely, `Open-PodeRunspace`). I'd also say to remove the `DisableRunspaceNaming` switches, if people want to change the name from the default standardised one - for say their schedule - they can do in their own ScriptBlock via `Set-PodeCurrentRunspaceName` even if it was already called prior, so I'm not sure what benefit the Disabling switch is adding.

This way, all Pode Runspaces get a default standardised name and if people need/want to set a different name they have that capability available to them as well. I'm happy with the setting of the primary `PodeServer` runspace name as well :)

I would also be very careful with Timers. The main Timer scheduler runspace will be named, but the runspace rename will change the scheduler runspace name; so the rapid renaming might get confusing, and it would never rename back to the original "scheduler" name so would look like one timer never ended.

---

I'm thinking the internal naming structure some be something like `Pode_<Type>_<Name>_<ID>`, where `<ID>` would be the ThreadId if available or just defaulted to 1. For example:

* Pode_Web_Listener_1
* Pode_Signals_Broadcaster_1
* Pode_Signals_Listener_1
* Pode_Web_KeepAlive_1
* Pode_Files_Watcher_1
* Pode_Main_Logging_1
* Pode_Timers_Scheduler_1
* Pode_Schedules_[Schedule Name]_1
* Pode_Tasks_[Task Name]_1

The `Add-PodeRunspace` function would need to have a new `-Name` and `-Id = 1` parameters added, as we generate the structured Runspace name near the top:
```powershell
$rsName = "Pode_$($Type)_$($Name)_$($Id)"
```

Where `Open-PodeRunspace` is called would have to change slightly, from:
```powershell
if (!$NoProfile) {
    $null = $ps.AddScript("Open-PodeRunspace -Type '$($Type)'")
}
```

to the following, where we pass the Name and NoProfile switch - removing the NoProfile `if` wrapper:
```powershell
$null = $ps.AddScript("Open-PodeRunspace -Type '$($Type)' -Name '$($rsName)' -NoProfile:`$$($NoProfile.IsPresent)")
```

Add then the `Open-PodeRunspace` would need a `-Name` and `-NoProfile` parameters, and be changed to:
```powershell
try {
    # set runspace name
    Set-PodeCurrentRunspaceName -Name $Name

    if (!$NoProfile) {
        # Importing internal Pode modules necessary for the runspace operations.
        Import-PodeModulesInternal

        # Adding PowerShell drives required by the runspace.
        Add-PodePSDrivesInternal
    }

    # Setting the state of the runspace pool to 'Ready', indicating it is ready to process requests.
    $PodeContext.RunspacePools[$Type].State = 'Ready'
}
catch {
    # ... unchanged ...
}
```

This would change, for example, the web listener runspace creation to:
```powershell
1..$PodeContext.Threads.General | ForEach-Object {
    Add-PodeRunspace -Type Web -Name 'Listener' -Id $_ -ScriptBlock $listenScript -Parameters @{ 'Listener' = $listener; 'ThreadId' = $_ }
}
```

Or for Logging:
```powershell
Add-PodeRunspace -Type Main -Name 'Logging' -ScriptBlock $script
```

Or when a Schedule triggers in `Invoke-PodeInternalScheduleLogic`:
```powershell
$runspace = Add-PodeRunspace -Type Schedules -Name $Schedule.Name -ScriptBlock $scriptblock -Parameters $parameters -PassThru
```

mdaneri:
I like that. 
I was thinking to do something similar but at the end I decided for something less invasive. But I think the way you are suggesting is great.

mdaneri:
Done but there is a bug in your last commit bug #1395

src/Private/FileWatchers.ps1
```diff
@@ -54,11 +54,7 @@ function Start-PodeFileWatcherRunspace {
     $watchScript = {
         param(
             [Parameter(Mandatory = $true)]
-            $Watcher,
-
-            [Parameter(Mandatory = $true)]
-            [int]
-            $ThreadId
```
Badgerati, line 61: The parameter needs to stay, plus the property from `Add-PodeRunspace -Parameters`, as I know of some people who do utilise this in their scripts to see which "thread" their script is being invoked from, and even `Write-PodeErrorLog` uses it 😃. Same in:

* PodeServer.ps1
* SmtpServer.ps1
* TcpServer.ps1
* Private/WebSockets.ps1

src/Private/Schedules.ps1
```diff
@@ -105,7 +106,7 @@ function Start-PodeScheduleRunspace {
         }
     }
 
-    Add-PodeRunspace -Type Main -ScriptBlock $script -NoProfile
+    Add-PodeRunspace -Type Main -Name 'ScheduleHouseKeeper' -ScriptBlock $script -NoProfile
```
Badgerati, line 109: This runspace isn't for the housekeeper, as that's further up as a Timer. This is for the scheduler for Schedules, so we could either call it `ScheduleScheduler` (😂), or just `Schedules`.

src/Private/WebSockets.ps1
```diff
@@ -107,7 +103,7 @@ function Start-PodeWebSocketRunspace {
 
     # start the runspace for listening on x-number of threads
     1..$PodeContext.Threads.WebSockets | ForEach-Object {
-        Add-PodeRunspace -Type WebSockets -ScriptBlock $receiveScript -Parameters @{ 'Receiver' = $PodeContext.Server.WebSockets.Receiver; 'ThreadId' = $_ }
+        Add-PodeRunspace -Type WebSockets -Name 'Listener' -Id $_ -ScriptBlock $receiveScript -Parameters @{ 'Receiver' = $PodeContext.Server.WebSockets.Receiver }
```
Badgerati, line 106: `-Name` here should be `Receiver` instead of Listener

docs/Getting-Started/Debug.md
```diff
@@ -174,3 +174,75 @@ The steps to attach to the Pode process are as follows:
...
+
+However, if a different runspace name is needed, Pode allows you to customize it. Inside the script block of `Add-PodeTask`, `Add-PodeSchedule`, or `Add-PodeTimer`, you can use the `Set-PodeCurrentRunspaceName` cmdlet to assign any custom name you prefer.
+
+```powershell
+Set-PodeCurrentRunspaceName -Name 'Custom Runspace Name'
+```
+
+This cmdlet sets a custom name for the runspace, making it easier to track during execution.
+
+### Example
```
Badgerati, line 212: This should be `####`

docs/Getting-Started/Debug.md
```diff
@@ -174,3 +174,75 @@ The steps to attach to the Pode process are as follows:
...
+
+Pode also provides the `Get-PodeCurrentRunspaceName` cmdlet to retrieve the name of the current runspace. This is particularly helpful when you need to log or display the runspace name dynamically during execution.
+
+```powershell
+Get-PodeCurrentRunspaceName
+```
+
+This cmdlet returns the name of the current runspace, allowing for easier tracking and management in complex scenarios with multiple concurrent runspaces.
+
+### Example
```
Badgerati, line 238: This should be `####`

docs/Getting-Started/Debug.md
```diff
@@ -174,3 +174,75 @@ The steps to attach to the Pode process are as follows:
 5. You'll also be able to query variables as well, such as `$WebEvent` and other variables you might have created.
 
 6. When you are done debugging the current request, hit the `d` key.
+
+
+
+## Customizing and Managing Runspace Names in Pode
```
Badgerati, line 180: I'd call this `## Managing Runspace Names`

docs/Getting-Started/Debug.md
```diff
@@ -174,3 +174,75 @@ The steps to attach to the Pode process are as follows:
...
+
+### Distinguishing Runspace Names in Pode
```
Badgerati, line 182: I'd call this `### Internal Runspace Naming`

Badgerati (changes requested):

src/Private/FileWatchers.ps1
mdaneri: I removed them because I thought that I added them to make the naming work.
I'm going to rollback them 😣

src/Private/Schedules.ps1
mdaneri: You are right I missed it

src/Private/WebSockets.ps1
mdaneri: Thanks here I was a bit confused between the roles. Comments were not very useful

Badgerati (approved):

PR #1344, opened 2024-06-27
Support passing Arrays to Functions Using Piping

mdaneri:
## Description

This feature enhances the Pode module by enabling full support for passing arrays to functions using the pipeline. The update ensures that the specified functions can correctly handle multiple input objects when piped as arrays, improving flexibility and usability in scripting and automation.

**Impacted Functions:**
- `Set-PodeCache`
- `Add-PodeOAResponse`
- `Remove-PodeOAResponse`
- `Set-PodeOARequest`
- `Set-PodeOARouteInfo`
- `Add-PodeOACallBack`
- `Add-PodeOAExternalRoute`
- `ConvertTo-PodeRoute`
- `Update-PodeSecret`
- `Set-PodeSecret`
- `Set-PodeState`
- `Out-PodeHost`
- `Write-PodeHost`
- `Out-PodeVariable`

**Key Changes:**
- Modified the impacted functions to support array input via the pipeline.
- Enhanced processing logic to aggregate piped values and handle them appropriately.
- Ensured backward compatibility by retaining existing parameter functionalities.

This update will streamline the process of handling multiple objects in a single pipeline, making it easier to work with collections of items in Pode scripts.

This update address #1343

Issue #1343, opened 2024-06-27
Set-PodeCache not working as expected with piped array


AdeeelAsif:
### Describe the Bug
`Set-PodeCache`  is not working as expected with a piped array. When I try to retrieve the cache value using `Get-PodeCache`, it only shows me the last value of the array, not the entire array.

### Steps To Reproduce

```
$test = $backendAddressPools | ForEach-Object {

                $fqdn = $.BackendAddresses.Fqdn
                $namespace = "Prd1" # Par défaut

                $health = $backendHealth | Where-Object { $.Address -eq $fqdn } | Select-Object -ExpandProperty Health

                if ($health -notmatch "Unknown") {

                    [PSCustomObject]@{
                        AppGWName         = $AppGwName
                        BackendName       = $.Name
                        provisioningstate = $.ProvisioningState
                        fqdn              = $_.BackendAddresses.Fqdn
                        namespace         = $namespace
                        health            = $health
                        date              = get-date
                    }
                }
            }

$test | Set-PodeCache -Key "AppGateway.$($AppGwName)" -Ttl 300
Write-PodeJsonResponse -Value @($test)
```
The problem does not occur when using `-InputObject ` instead of piping the array.

Badgerati:
Resolved via #1388

PR #1388, opened 2024-09-08
Support passing Arrays to Functions Using Piping

mdaneri:
recovered from Support passing Arrays to Functions Using Piping #1344
## Description

This feature enhances the Pode module by enabling full support for passing arrays to functions using the pipeline. The update ensures that the specified functions can correctly handle multiple input objects when piped as arrays, improving flexibility and usability in scripting and automation.

**Impacted Functions:**
- `Set-PodeCache`
- `Add-PodeOAResponse`
- `Remove-PodeOAResponse`
- `Set-PodeOARequest`
- `Set-PodeOARouteInfo`
- `Add-PodeOACallBack`
- `Add-PodeOAExternalRoute`
- `ConvertTo-PodeRoute`
- `Update-PodeSecret`
- `Set-PodeSecret`
- `Set-PodeState`
- `Out-PodeHost`
- `Write-PodeHost`
- `Out-PodeVariable`

**Key Changes:**
- Modified the impacted functions to support array input via the pipeline.
- Enhanced processing logic to aggregate piped values and handle them appropriately.
- Ensured backward compatibility by retaining existing parameter functionalities.

This update will streamline the process of handling multiple objects in a single pipeline, making it easier to work with collections of items in Pode scripts.

This update address #1343


src/Private/Helpers.ps1
```diff
@@ -1415,7 +1442,7 @@ function New-PodeRequestException {
 
 function ConvertTo-PodeResponseContent {
```
Badgerati, line 1443: There's quite a few of these internal functions where they have parameters flagged with `ValueFromPipeline = $true`, but I never actually use the pipeline - mostly to help with performance, and directly pass the object instead.

`ConvertTo-PodeResponseContent` and `ConvertFrom-PodeHeaderQValue` are two such examples, but there are others above and below which I've crossed.

I'm thinking we just remove the attribute instead, from parameters on functions where the values just aren't being passed via pipeline.

src/Private/Responses.ps1
```diff
@@ -133,7 +133,7 @@ This is an internal function and may change in future releases of Pode.
 function Write-PodeFileResponseInternal {
     [CmdletBinding()]
     param (
-        [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+        [Parameter(Mandatory = $true, Position = 0, ValueFromPipeline = $true)]
```
Badgerati, line 136: This would be another example of one where we could simply remove the ValueFromPipeline attribute instead; it's never used, and means we don't need to alter the function's logic brlow.

Badgerati (changes requested):
I've reviewed some, but two comments reference examples where we could simplify further by just removing all unneeded `ValueFromPipeline` attributes from Private functions - doing this for quite a few of the functions will reduce a lot of the changes as they wouldn't be needed.

src/Private/Helpers.ps1
```diff
@@ -927,12 +927,25 @@ function Remove-PodeEmptyItemsFromArray {
...
+        if ($null -eq $Array) {
+            return @()
+        }
 
-    return @( @($Array -ne ([string]::Empty)) -ne $null )
+        return @( @($Array -ne ([string]::Empty)) -ne $null )
+    }
 }
 
 function Remove-PodeNullKeysFromHashtable {
```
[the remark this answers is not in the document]
mdaneri: I agree

src/Private/Responses.ps1
mdaneri: I'm fine with that. But we are breaking from the previous versions

src/Private/Helpers.ps1
mdaneri: are internal function we can do it

src/Private/Helpers.ps1
```diff
@@ -927,12 +927,25 @@ function Remove-PodeEmptyItemsFromArray {
...
+        if ($pipelineValue.Count -gt 1) {
+            $Array = $pipelineValue
+        }
+        if ($null -eq $Array) {
+            return @()
+        }
 
-    return @( @($Array -ne ([string]::Empty)) -ne $null )
+        return @( @($Array -ne ([string]::Empty)) -ne $null )
+    }
```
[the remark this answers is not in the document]
mdaneri: Much better I removed the piping completely. Whe never use it

src/Public/Utilities.ps1
```diff
@@ -772,14 +796,34 @@ The object to output.
...
+            return
+        }
+        # Set InputObject to the array of values
+        if ($pipelineValue.Count -gt 1) {
+            $InputObject = $pipelineValue
+            $InputObject | Out-Default
+        }
+        else {
+            Out-Default -InputObject $InputObject
+        }
```
Badgerati, line 824: How come not just the following?
```powershell
if ($pipelineValue.Count -gt 1) {
    $InputObject = $pipelineValue
}

$InputObject | Out-Default
```

Badgerati (changes requested):

src/Public/SSE.ps1
```diff
@@ -235,50 +235,64 @@ function Send-PodeSseEvent {
...
+            return
         }
     }
 
-    # send directly back to current connection
-    if ($FromEvent -and $WebEvent.Sse.IsLocal) {
-        $null = Wait-PodeTask -Task $WebEvent.Response.SendSseEvent($EventType, $Data, $Id)
-        return
+    process {
+        if ($PSCmdlet.ParameterSetName -eq 'Value') {
```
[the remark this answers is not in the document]
mdaneri: because I forgot to port the modified version from AsyncRoute

src/Public/Secrets.ps1
```diff
@@ -660,50 +660,65 @@ function Update-PodeSecret {
...
+    end {
+        # Set InputObject to the array of values
+        if ($pipelineValue.Count -gt 1) {
+            $InputObject = $pipelineValue
+        }
 
-    # get the secret and vault
-    $secret = $PodeContext.Server.Secrets.Keys[$Name]
+        # make sure the value type is correct
+        $InputObject = Protect-PodeSecretValueType -Value $InputObject
```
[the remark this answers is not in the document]
mdaneri: you are right regarding `Protect-PodeSecretValueType` but I can change it to make work

Badgerati (approved):

Issue #1412, opened 2024-10-10
Test-PodeCacheStorage called with incorrect param


willgladstone:
### Describe the Bug
When using Redis as external cache storage, the below error is thrown
```Test-PodeCacheStorage: ~/.local/share/powershell/Modules/Pode/2.11.0/Public/Caching.ps1:153
Line |
 153 |          elseif (Test-PodeCacheStorage -Key $Storage) {
     |                                        ~~~~
     | A parameter cannot be found that matches parameter name 'Key'.```

### Problematic Code References
https://github.com/Badgerati/Pode/blob/21c76a16c0b57caa459fbc9de4a7d13b412133c1/src/Public/Caching.ps1#L55
https://github.com/Badgerati/Pode/blob/21c76a16c0b57caa459fbc9de4a7d13b412133c1/src/Public/Caching.ps1#L153
https://github.com/Badgerati/Pode/blob/21c76a16c0b57caa459fbc9de4a7d13b412133c1/src/Public/Caching.ps1#L207
https://github.com/Badgerati/Pode/blob/21c76a16c0b57caa459fbc9de4a7d13b412133c1/src/Public/Caching.ps1#L258
https://github.com/Badgerati/Pode/blob/21c76a16c0b57caa459fbc9de4a7d13b412133c1/src/Public/Caching.ps1#L304


### Expected Behavior
The Test-PodeCacheStorage expects a `Name` parameter.

### Additional Context
Not sure of the preferred fixed here. The least intrusive is to set an alias, or the params could be fixed to used the correct parameter name.

Badgerati:
Fixed via #1413

PR #1413, opened 2024-10-10
Fixes Test-PodeCacheStorage being called with incorrect -Key parameter

willgladstone:
Addresses https://github.com/Badgerati/Pode/issues/1412

### DIFF
--- a/src/Public/Caching.ps1
+++ b/src/Public/Caching.ps1
@@ -462,6 +462,7 @@ if (Test-PodeCacheStorage -Name 'ExampleStorage') { }
 function Test-PodeCacheStorage {
     [CmdletBinding()]
     param(
+        [Alias('Key')]
         [Parameter(Mandatory = $true)]
         [string]
         $Name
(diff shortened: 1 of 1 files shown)


src/Public/Caching.ps1
```diff
@@ -462,6 +462,7 @@ if (Test-PodeCacheStorage -Name 'ExampleStorage') { }
 function Test-PodeCacheStorage {
     [CmdletBinding()]
     param(
+        [Alias('Key')]
         [Parameter(Mandatory = $true)]
         [string]
         $Name
```
Badgerati, line 468: `-Name` is the correct parameter here for `Test-PodeCacheStorage`. In the locations where `Test-PodeCacheStorage` is being erroneously called with `-Key` instead of `-Name`, those need to be updated to be `-Name`, rather than adding an Alias.

Badgerati (changes requested):

Badgerati (approved):
LGTM

Issue #1417, opened 2024-10-15
Issue with `ConvertTo-PodeYaml` when `Count` is a Key in Hashtable or PSCustomObject


mdaneri:
#### **Summary:**
There is an issue in the `ConvertTo-PodeYaml` function in Pode that causes the conversion to fail if the source `Hashtable` or `PSCustomObject` contains a key named `Count`. This problem occurs due to the way the `Count` property is being checked during conversion, leading to conflicts when `Count` is a key in the object.

#### **File:**
- **File Name:** `Helper.ps1`
- **Location:** 
  - Line 3576: `if ($InputObject.Count -gt 0) {`
  - Line 3602: `if ($InputObject.Count -gt 0) {`

#### **Steps to Reproduce:**
1. Create a `Hashtable` or `PSCustomObject` with a key named `Count`.
2. Attempt to convert the object to YAML using `ConvertTo-PodeYaml`.
3. The conversion will fail due to a conflict with the `Count` property being treated as a method rather than a key.

#### **Example:**
```powershell
$object = @{
    Count = 5
    Name = "TestObject"
}

ConvertTo-PodeYaml -InputObject $object
```

#### **Root Cause:**
The issue arises because the `Count` property is being accessed directly (`$InputObject.Count`), which conflicts when `Count` is used as a key in the `Hashtable` or `PSCustomObject`. The `Count` property is treated as a method in such cases, causing the conversion to fail.

#### **Fix:**
The proposed solution is to adjust the conditions to avoid using the direct `Count` property in situations where `Count` could be a key.

1. **Line 3576:**
   - **Current Code:**
     ```powershell
     if ($InputObject.Count -gt 0) {
     ```
   - **Proposed Fix:**
     ```powershell
     if ($hashtable.GetEnumerator().MoveNext()) {
     ```

2. **Line 3602:**
   - **Current Code:**
     ```powershell
     if ($InputObject.Count -gt 0) {
     ```
   - **Proposed Fix:**
     ```powershell
     if ($InputObject.PSObject.Properties.Count -gt 0) {
     ```

#### **Additional Notes:**
- The fix ensures that when `Count` is used as a key, it will not conflict with the `Count` property of the object, and the conversion to YAML will work as expected.
- The proposed fix uses `.GetEnumerator().MoveNext()` to check for the presence of elements in a `Hashtable`, and `.PSObject.Properties.Count` to properly count the properties in a `PSCustomObject`.

PR #1418, opened 2024-10-15
Fix for ConvertTo-PodeYaml failing, when passed object contains a Key named "Count"

mdaneri:
#### **Issue Reference**
This pull request addresses the issue described in [Bug #1417](https://github.com/Badgerati/Pode/issues/1417), where `ConvertTo-PodeYaml` fails when the source `Hashtable` or `PSCustomObject` contains a key named `'Count'`.

#### **Issue Description**
The `ConvertTo-PodeYaml` function incorrectly assumes that `.Count` refers to the number of elements in the object, which causes a conflict if the object contains a key named `'Count'`. This results in a failure during the YAML conversion process.

#### **Affected File**
- **File**: `Helper.ps1`
- **Lines**:
  - **Line 3576**: `if ($InputObject.Count -gt 0 )`
  - **Line 3602**: `if ($InputObject.Count -gt 0 )`

#### **Proposed Fix**
To avoid conflicts with the `'Count'` key, the following changes have been made:
- **Line 3576**: Replaced with `if ($InputObject.GetEnumerator().MoveNext())`
- **Line 3602**: Replaced with `if ($InputObject.PSObject.Properties.Count -gt 0)`

#### **Code Changes**
```diff
- if ($InputObject.Count -gt 0 ) {
+ if ($InputObject.GetEnumerator().MoveNext()) {

- if ($InputObject.Count -gt 0 ) {
+ if ($InputObject.PSObject.Properties.Count -gt 0) {
```

#### **Testing**
The issue has been reproduced with a `Hashtable` and `PSCustomObject` containing a key named `'Count'`. After applying the fix, the `ConvertTo-PodeYaml` function works as expected without any errors.

#### **Example of Reproducing the Issue**
Before the fix:
```powershell
$data = @{
    Name  = 'Sample'
    Count = 10
}
ConvertTo-PodeYaml -InputObject $data
```
This would throw an error due to the key `'Count'`.

After the fix, the same command should successfully convert the object to YAML.


Badgerati (approved):
LGTM

PR #1428, opened 2024-10-22
Migrates Stream functions into the .NET Listener

### LINKED ISSUE #1426: Rewrite Steams.ps1 into .NET, to prevent VirusTotal erroneously flagging the file
### Describe the Change
This was an issue found while releasing 2.11.0 to Chocolatey (Choco). Since they run scans on packages via VirusTotal (VT), it would appear that some vendors are erroneously flagging the `Private/Streams.ps1` file. The flagging is due to the file containing many functions which are for reading, altering, and writing to streams - which is seen as "potentially malicious" even though it's been used in perfectly valid context.

You can see an example here: https://www.virustotal.com/gui/file/1e89d36213e613f709588bf253193574bc47fbf4a797223a44ab7742a924676f

To save having to communicate with 10 different vendors, and allow Choco to publish the package without exemptions, a quick fix here could be as simple as moving the functions into the compiled .NET DLL - which could even help give a quick perf boost.

When the rewrite is done, a quick adhoc packaging locally and running the package via VT should show 0 violations.

Badgerati:
### Description of the Change
Migrates PowerShell streaming functions into the compiled .NET DLL, to prevent VirusTotal from flagging the `Private/Streams.ps1` file as "malicious" just because it reading and compresses streams!

### Related Issue
Resolves #1426 

### Additional Context
Scan results after changes
image


mdaneri:
it's not part of your pull request but in PodeHelper line 143 
```c#
if (count > 0)
{
       await stream.WriteAsync(array, startIndex, count, cancellationToken).ConfigureAwait(false);
}
```

can be improved performance-wise using

```c#
if (count > 0)
{
      await stream.WriteAsync(array.AsMemory(startIndex, count), cancellationToken).ConfigureAwait(false);
}
```


https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835

Badgerati:
This one I'm aware of, but as Spans/Memory were .NET8 only I didn't include them originally. I have thought about wrapping the code in `#if NET8_0_OR_GREATER` for the various places that would benefit from it though

PR #1449, opened 2024-11-06
Change netstandard2.0 to net48 (test)

mdaneri:
### Description of the Change
Change netstandard2.0 to net48

### DIFF
--- a/.github/workflows/VirusTotal-Releases.yml
+++ b/.github/workflows/VirusTotal-Releases.yml
@@ -9,6 +9,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Check VIRUSTOTAL_API_KEY
+      env:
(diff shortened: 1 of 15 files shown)


Badgerati:
Just did some testing myself, as I had concerns about how this might affect people on Linux and PS7.0-7.2 - more so as PS7.2 is EOL this month. Seems to work, which is interesting!

Couple of questions:
1. Any reason not to just bump this to `net481`? Be the most up-to-date then, and still worked in my testing as well
2. Seeing as 2.12.0 will be going live after PS7.2 and NET6 go EOL, we probably want to remove the compiling of `net6` as well. Either as a different PR, or just a part of this will do?

mdaneri:
regarding net481 I was reading that is not working on 5.1. I haven't tried yet

Issue #1459, opened 2024-11-29
ObjectDisposedException in PodeRequest During SslStream Operation


mdaneri:
#### **Summary**
An `ObjectDisposedException` is thrown intermittently during SSL/TLS operations when running a Pode server using the `Start-PodeServer` function. The exception occurs in the `PodeRequest.Receive` method while handling an `SslStream`.

#### **Steps to Reproduce**
1. Start a Pode server with the following configuration:
   ```powershell
   Start-PodeServer -Threads 6 {
       Add-PodeEndpoint -Address "*" -Port "8080" -Protocol "Https" -SelfSigned

       New-PodeLoggingMethod -File -Name "requests" | Enable-PodeRequestLogging
       New-PodeLoggingMethod -File -Name "errors" | Enable-PodeErrorLogging

       Add-PodeRoute -Method Get -Path "/" -ScriptBlock {
           Write-PodeTextResponse -Value "Test"
       }
   }
   ```

2. Send multiple HTTPS requests to the server simultaneously (e.g., using a browser or a load-testing tool).

#### **Observed Behavior**
The server intermittently logs the following error:
```
[Error] ObjectDisposedException: Cannot access a disposed object.
Object name: 'SslStream'.
   at System.Net.Security.SslState.CheckThrow(Boolean authSuccessCheck, Boolean shutdownCheck)
   at System.Net.Security.SslState.get_SecureStream()
   at System.Net.Security.SslStream.EndRead(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrimPromise`1.Complete(TInstance thisRef, Func`3 endMethod, IAsyncResult asyncResult, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Pode.PodeRequest.<Receive>d__83.MoveNext()
```

#### **Expected Behavior**
The Pode server should process HTTPS requests without encountering an `ObjectDisposedException`.

#### **Potential Cause**
The issue appears to be related to concurrent access or premature disposal of the `SslStream` instance. It may occur under high concurrency due to:
1. Multiple threads attempting to access the same `SslStream` object.
2. The stream being disposed of while another operation is still in progress.

#### **Relevant Code**
- The issue is observed in the `PodeRequest.Receive` method, specifically during operations involving `SslStream`.
[...]

PR #1460, opened 2024-11-29
Fixes ObjectDisposedException in during SSL/TLS operations

mdaneri:
#### **Summary**
This pull request addresses #1459 
Intermittent `ObjectDisposedException` that occurs in the `PodeRequest.Receive` method during SSL/TLS operations in Pode. The issue arises due to premature disposal or concurrent access of the `SslStream` instance when handling HTTPS requests under high concurrency.

#### **Problem**
The following error was observed when running a Pode server with an HTTPS endpoint under load:

```
[Error] ObjectDisposedException: Cannot access a disposed object.
Object name: 'SslStream'.
   at System.Net.Security.SslState.CheckThrow(Boolean authSuccessCheck, Boolean shutdownCheck)
   at System.Net.Security.SslState.get_SecureStream()
   at System.Net.Security.SslStream.EndRead(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncTrimPromise`1.Complete(TInstance thisRef, Func`3 endMethod, IAsyncResult asyncResult, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Pode.PodeRequest.<Receive>d__83.MoveNext()
```

The issue is caused by:
1. **Premature Disposal**: The `SslStream` instance is disposed of while operations are still in progress.
2. **Concurrent Access**: Multiple threads attempting to access the `SslStream` without proper synchronization.

#### **Changes Made**
1. **Improved Thread Safety**:
   - Added additional checks to ensure that all operations on `SslStream` are thread-safe, using `SemaphoreSlim` to manage concurrency.

2. **Enhanced Disposal Logic**:
   - Updated the `Dispose` method in `PodeRequest` to ensure that the `SslStream` is not disposed prematurely.
   - Introduced `IsDisposed` and `IsOpen` state checks to prevent operations on already-disposed streams.

3. **Error Logging**:
   - Adjusted error logging levels for expected exceptions (`ObjectDisposedException` and `IOException`) to avoid alarming users unnecessarily. These exceptions are now logged at the `Verbose` level instead of `Error`.

4. **Buffer Management**:
   - Optimized buffer handling to prevent unnecessary reinitializations and ensure reuse during stream operations.

5. **Stream Cleanup**:
   - Ensured proper handling and disposal of streams, including defensive null checks and resetting state variables like `InputStream`.

#### **Results**
- The `ObjectDisposedException` no longer occurs under high-concurrency scenarios.
- The Pode server handles concurrent HTTPS requests more reliably without resource contention or premature disposal.

#### **Impact**
- Improved stability and reliability for HTTPS endpoints in Pode.
- Resolved a critical issue affecting high-concurrency scenarios, ensuring seamless handling of SSL/TLS operations.

#### **Backward Compatibility**
- These changes are fully backward-compatible with existing Pode functionality.


src/Listener/PodeRequest.cs
```diff
@@ -292,108 +316,23 @@ public async Task<bool> Receive(CancellationToken cancellationToken)
                 PodeHelpers.WriteException(ex, Context.Listener, PodeLoggingLevel.Error);
                 Error = ex;
             }
+            catch (ObjectDisposedException ex)
+            {
+                PodeHelpers.WriteException(ex, Context.Listener, PodeLoggingLevel.Error);
+            }
```
Badgerati, line 322: I would just let this fall into the `catch (Exception ex)` block below, as that has the same logging but also sets the status code to 500.

src/Listener/PodeRequest.cs
```diff
@@ -292,108 +316,23 @@ public async Task<bool> Receive(CancellationToken cancellationToken)
...
             return false;
         }
 
-        /// <summary>
-        /// Reads data from the input stream until the specified bytes are found.
-        /// </summary>
-        /// <param name="checkBytes">The bytes to check for in the input stream.</param>
-        /// <param name="cancellationToken">Token to monitor for cancellation requests.</param>
-        /// <returns>A Task representing the async operation, with a string containing the data read.</returns>
-        public async Task<string> Read(byte[] checkBytes, CancellationToken cancellationToken)
```
Badgerati, line 314: Could we re-add this `Read` method, and the `ValidateInputInternal` method below please? I know Read isn't used currently, but I will need it eventually for another feature I'm looking to add.

src/Listener/PodeRequest.cs
```diff
@@ -421,41 +360,78 @@ protected virtual bool ValidateInput(byte[] bytes)
...
+            }
         }
 
         /// <summary>
-        /// Disposes of the request and its associated resources.
+        /// Dispose managed and unmanaged resources.
         /// </summary>
-        public virtual void Dispose()
+        /// <param name="disposing">Indicates if disposing is called manually or by garbage collection.</param>
+        protected virtual void Dispose(bool disposing)
```
Badgerati, line 388: I don't see the need for having `disposing` and the deconstructor below?

If GC comes along, surely we want it to dispose of the Stream/Socket? And in the logic below, even if `disposing=false` we still set `IsDisposed=true` anyway 🤔 

It's also better for `IsDisposed=true` to be set further up, before we actually dispose of objects, so we're not disposing of objects and then finally setting it to true - which does open the possibility for `ObjectDisposedException` as `Receive` could use them

src/Listener/PodeHttpRequest.cs
```diff
@@ -412,22 +412,33 @@ public override void PartialDispose()
...
+                {
+                    BodyStream.Dispose();
+                    BodyStream = null;
+                }
+
+                if (Form != null)
+                {
+                    Form.Dispose();
+                    Form = null;
+                }
```
Badgerati, line 437: Why the change from `default(...)`?

Badgerati (changes requested):

src/Listener/PodeHttpRequest.cs
mdaneri: because I was lazy :)

src/Listener/PodeRequest.cs
mdaneri: You're absolutely right to question the need for both the destructor and the disposing = false logic. Since the class only deals with managed resources like streams and sockets, the destructor (~PodeRequest) isn't necessary. 
Removing it simplifies things and avoids the disposing = false scenario altogether.
As for setting IsDisposed = true earlier—great point! Doing that upfront helps prevent potential ObjectDisposedException during concurrent use by marking the object as disposed before starting cleanup.
😊

mdaneri: regarding why to use `Dispose(bool disposing)`
https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

src/Listener/PodeRequest.cs
```diff
@@ -221,42 +237,50 @@ private bool ValidateCertificateCallback(object sender, X509Certificate certific
         /// <returns>A Task representing the async operation, with a boolean indicating whether the connection should be closed.</returns>
         public async Task<bool> Receive(CancellationToken cancellationToken)
         {
-            // Check if the stream is open
-            if (State != PodeStreamState.Open)
-            {
-                return false;
-            }
-
+            await StreamLock.WaitAsync(cancellationToken);
```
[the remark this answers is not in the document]
mdaneri: yes you are right. I added it when I was trying to fix the reported issue.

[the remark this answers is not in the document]
mdaneri: I have to reinstate it because otherwise everything fails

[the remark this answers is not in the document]
Badgerati: For this, I was more referring that the `StreamLock` variable entirely can go - because of how code will drop into this method, Semaphores shouldn't be required, and the locking will just add overhead.

src/Listener/PodeRequest.cs
Badgerati: Aaah, never knew about the hidden `dispose(bool)`. If I'm reading it correctly then in, for example, `PodeHttpRequest` the `RawBody` and `_body` variables being nullied should be outside of the `if (disposing)` since they're unmanaged, while `BodyStream`/`Form`/etc should be within the `if` because they're managed? 🤔

mdaneri: it's not a requirement. All resources are disposed at the end doesn't matter if best practice are followed or not

src/Listener/PodeRequest.cs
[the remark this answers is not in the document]
mdaneri: I cannot remove the Semaphores. there are multiple receive calls I presume.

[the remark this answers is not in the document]
mdaneri: I added some output and this is the result 
[Receive][Receive][Close:False]<end_block><finally><end_finally>
[Receive][Close:False]<end_block><finally><end_finally>
[Receive][Receive]<OperationCanceledException><finally><end_finally>
<OperationCanceledException><finally><OperationCanceledException><finally>[Receive][Close:False]<end_block><finally><end_finally>
[Receive][Receive]<OperationCanceledException><finally><end_finally>
<OperationCanceledException><finally>

multiple receive requests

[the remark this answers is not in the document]
mdaneri: My bad I forgot to remove the semaphore from the dispose. That was the reason of the issue

src/Listener/PodeRequest.cs
```diff
@@ -145,56 +146,70 @@ public async Task Open(CancellationToken cancellationToken)
...
-                {
-                    PodeHelpers.WriteException(ex, Context.Listener, PodeLoggingLevel.Debug);
-                    State = PodeStreamState.Error;
-                    Error = new PodeRequestException(ex, 400);
-                }
-                catch (Exception ex)
+                    if (ssl != null && !IsDisposed)
+                    {
+                        ssl.Dispose();
+                    }
```
Badgerati, line 174: Could be simplified as this?

```c#
if (!IsDisposed)
{
    ssl?.Dispose();
}
```

src/Listener/PodeRequest.cs
```diff
@@ -421,41 +443,62 @@ protected virtual bool ValidateInput(byte[] bytes)
...
-            }
+                if (InputStream != default(Stream))
+                {
+                    InputStream.Dispose();
+                    InputStream = default;
+                }
 
-            if (InputStream != default(Stream))
-            {
-                State = PodeStreamState.Closed;
```
Badgerati, line 452: Just spotted that `State = PodeStreamState.Closed` has been removed, this needs to be added back otherwise some of the logic which checks State will still think the Request is open.

Badgerati (changes requested):

src/Listener/PodeRequest.cs
```diff
@@ -39,19 +37,17 @@ public class PodeRequest : PodeProtocol, IDisposable
         public X509Certificate Certificate { get; private set; }
         public bool AllowClientCertificate { get; private set; }
         public PodeTlsMode TlsMode { get; private set; }
-        public X509Certificate2 ClientCertificate { get; set; }
-        public SslPolicyErrors ClientCertificateErrors { get; set; }
+        public X509Certificate2 ClientCertificate { get; private set; }
+        public SslPolicyErrors ClientCertificateErrors { get; private set; }
```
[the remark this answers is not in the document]
mdaneri: I don't see this function anywhere. and don't see where this property is used. But if you say so I'm going to change it

src/Listener/PodeRequest.cs
mdaneri: it's not the same but achieve the same goal. my version is faster your is cleaner :)

src/Listener/PodeRequest.cs
mdaneri: sorry I don't see it 
line 452:  State = PodeStreamState.Closed;

mdaneri: the merge that create this false positive

src/Listener/PodeRequest.cs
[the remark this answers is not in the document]
Badgerati: It's in Private/Middleware.ps1 line ~391

src/Listener/PodeRequest.cs
Badgerati: In the checked out PR code, the setting of State to Closed has been removed: 
image

src/Listener/PodeRequest.cs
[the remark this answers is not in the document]
mdaneri: It is strange that Visual Studio Code was unable to find it before. I believe there is a bug in VSC where frequent branch changes cause the search function to malfunction.

src/Listener/PodeRequest.cs
mdaneri: you are right sorry

Badgerati (approved):

PR #1469, opened 2025-01-28
Improved Access and Rate Limiting Middleware

### LINKED ISSUE #1462: debounce for get/post request
Hi 

I am wondering whether pode has debounce descriptor.

Thanks
Eric Xin

Badgerati:
### Description of the Change
This replaces the current `Add-PodeAccessRule` and `Add-PodeLimitRule`, which are too restrictive - as they only allow rate limiting to be enabled on just an IP, and nothing more complex such as IP/Route combinations. These older functions, while now deprecated, call the newer functions for backwards compatibility support.

The newer functions are `Add-PodeLimitRateRule` and `Add-PodeLimitAccessRule` - and they are also joined by their accompanying `Remove`, `Test`, and `Get` methods.

For the new functions, you "build" a rule using various components; currently inbuilt are IP, Route, Endpoint, Method, and Header - but the syntax also enables custom components to be built as well:

* `New-PodeLimitIPComponent`
* `New-PodeLimitRouteComponent`
* `New-PodeLimitEndpointComponent`
* `New-PodeLimitMethodComponent`
* `New-PodeLimitHeaderComponent`

Each component returns a hashtable, containing an "Options" property (arguments to pass to the scriptblock), and a "Scriptblock" property which verifies the request against the current component/options.

For example, a simple "Rate limit all IPs, so they can only do 10 requests per second":

```powershell
Add-PodeLimitRateRule -Name 'Example' -Limit 10 -Timeout 1 -Component @(
    New-PodeLimitIPComponent
)
```

Or, if we want to: "Rate limit IPs, so they can only hit separate GET/POST Routes, under /api/*, once per minute":

```powershell
Add-PodeLimitRateRule -Name 'Example' -Limit 1 -Timeout 60 -Component @(
    New-PodeLimitIPComponent
    New-PodeLimitRouteComponent -Path '/api/*'
    New-PodeLimitMethodComponent -Method Get, Post
)
```

The above will rate limit each route individually, if instead you wanted to apply a global base rate limit to all routes, as one entity, under /api/*:

```powershell
Add-PodeLimitRateRule -Name 'Example' -Limit 1 -Timeout 60 -Component @(
    New-PodeLimitIPComponent
    New-PodeLimitRouteComponent -Path '/api/*' -Group
    New-PodeLimitMethodComponent -Method Get, Post
)
```

Or, for access: "Block all requests from a subnet":

```powershell
Add-PodeLimitAccessRule -Name 'Example' -Action Deny -Component @(
    New-PodeLimitIPComponent -IP '10.0.0.0/24'
)
```

Or, "Block all requests, from a subnet, to /api/*":

```powershell
Add-PodeLimitAccessRule -Name 'Example' -Action Deny -Component @(
    New-PodeLimitIPComponent -IP '10.0.0.0/24'
    New-PodeLimitRouteComponent -Path '/api/*'
)
```

### Related Issue
Resolves #1462


mdaneri:
what do you think to  include the server disable feature in this new Limit feature?

 Current  `Disable-PodeServerInternal`
```
function Disable-PodeServerInternal {

    if (!(Test-PodeServerState -State Running) -or (!( Test-PodeServerIsEnabled)) ) {
        return
    }

    # Trigger the 'Enable' event for the server.
    Invoke-PodeEvent -Type Disable

    # Add middleware to block new requests and respond with 503 Service Unavailable
    Add-PodeMiddleware -Name  $PodeContext.Server.AllowedActions.DisableSettings.MiddlewareName -ScriptBlock {
        # Set HTTP response header for retrying after a certain time (RFC7231)
        Set-PodeHeader -Name 'Retry-After' -Value $PodeContext.Server.AllowedActions.DisableSettings.RetryAfter

        # Set HTTP status to 503 Service Unavailable
        Set-PodeResponseStatus -Code 503

        # Stop further processing
        return $false
    }
}
```

Badgerati:
We could actually 🤔 I'll have a look, might have an idea to to make it work with some small tweaks

Badgerati:
`Disable-PodeServer` now adds a rate limiting rule, instead of middleware.

Badgerati:
@mdaneri I've finished everything up here; if you're also happy this solves the debouncing request as well I'll get it merged.

Issue #1493, opened 2025-02-20
Additional information in the raw request hashtable for custom logging + UTC option


Badgerati:
### Discussed in https://github.com/Badgerati/Pode/discussions/1491

<div type='discussions-op-text'>

<sup>Originally posted by **chris-dma** February 18, 2025</sup>
Would it be possible to look at including additional information in the raw request hash table for custom request logging? Specifically, it would be great if we could have the hostname and the query strings portion of the request included. Additionally, it would be great if there was an option to use UTC in the raw hash table as well. 

Here is an example of what I mean:

@{
    Host = '10.10.0.3'
    RfcUserIdentity = '-'
    User = '-'
    Date = '2018-06-14T20:23:52Z'
    Request = @{
        Method = 'GET'
        Hostname = 'https://api.domain.com'
        Resource = '/api/users'
        QueryStrings = 'query1=vlaue&query2=value'
        Protocol = "HTTP/1.1"
        Referrer = '-'
        Agent = '<user-agent>'
    }
    Response = @{
        StatusCode = '200'
        StautsDescription = 'OK'
        Size = '9001'
    }
}

I am currently using a custom logging script block to change the date format to UTC and then convert the raw hash table to compressed JSON then writing that to a newline-delimited JSON logfile, which is then picked up by the Elastic Agent to ingest the request logs. This works quite well but it would save some processing if the date format could be in UTC format already and the additional information would be helpful to be included. 
</div>

Badgerati:
@chris-dma I can look at adding the Querystring and Hostname, the data is a little different as it's best to use `[DateTime]::Now` and have the actually underlying server configured to use UTC. However, I can see an enhancement with the dates being passed, where it'd be better to have the raw DateTime object and not a string - so I might have a `Date` and a `UtcDate` in the raw object

chris-dma:
@Badgerati thank you. I've had a look at the Logging.ps1 and create a pull request which adds these to the $item variable. Apologies if I have done anything wrong with the pull request - this is my first time using GitHub.

PR #1495, opened 2025-02-20
Adds Hostname, Scheme, Query, and UtcDate to raw Request item for logging

chris-dma:
### Description of the Change
Adds hostname, query, and UtcDate to the raw item for custom logging

### Related Issue
Fixes Badgerati/Pode#1493

### DIFF
--- a/src/Private/Logging.ps1
+++ b/src/Private/Logging.ps1
@@ -299,12 +299,13 @@ function Write-PodeRequestLog {
         RfcUserIdentity = '-'
         User            = '-'
         Date            = [DateTime]::Now.ToString('dd/MMM/yyyy:HH:mm:ss zzz')
-        UtcDate         = [DateTime]::Now.ToUniversalTime()
+        UtcDate         = [DateTime]::UtcNow
         Request         = @{
             Method   = $Request.HttpMethod.ToUpperInvariant()
-            Hostname = "$($Request.Scheme.ToLowerInvariant())://$($Request.Host.ToLowerInvariant())"
+            Hostname = $Request.Host.ToLowerInvariant()
+            Scheme   = $Request.Scheme.ToLowerInvariant()
             Resource = $Path
-            Query = '-'
+            Query = (Protect-PodeValue -Value $Request.Url.Query -Default '-').TrimStart('?')
             Protocol = "HTTP/$($Request.ProtocolVersion)"
             Referrer = $Request.UrlReferrer
             Agent    = $Request.UserAgent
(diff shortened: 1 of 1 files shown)


src/Private/Logging.ps1
```diff
@@ -299,9 +299,12 @@ function Write-PodeRequestLog {
         RfcUserIdentity = '-'
         User            = '-'
         Date            = [DateTime]::Now.ToString('dd/MMM/yyyy:HH:mm:ss zzz')
+        UtcDate         = [DateTime]::Now.ToUniversalTime()
```
Badgerati, line 302: This can just be `[DateTime]::UtcNow`

src/Private/Logging.ps1
```diff
@@ -299,9 +299,12 @@ function Write-PodeRequestLog {
...
         Request         = @{
             Method   = $Request.HttpMethod.ToUpperInvariant()
+            Hostname = "$($Request.Scheme.ToLowerInvariant())://$($Request.Host.ToLowerInvariant())"
             Resource = $Path
+            Query = '-'
```
Badgerati, line 307: To help replace the query logic below, you can simplify this by using `$Request.Url.Query` and `Protect-PodeValue` :)

```powershell
Query = (Protect-PodeValue -Value $Request.Url.Query -Default '-').TrimStart('?')
```

src/Private/Logging.ps1
```diff
@@ -299,9 +299,12 @@ function Write-PodeRequestLog {
         RfcUserIdentity = '-'
         User            = '-'
         Date            = [DateTime]::Now.ToString('dd/MMM/yyyy:HH:mm:ss zzz')
+        UtcDate         = [DateTime]::Now.ToUniversalTime()
         Request         = @{
             Method   = $Request.HttpMethod.ToUpperInvariant()
+            Hostname = "$($Request.Scheme.ToLowerInvariant())://$($Request.Host.ToLowerInvariant())"
```
Badgerati, line 305: Could we split this into two properties? Just so the hostname is *the* hostname, and the two can be concatenated when needed in log processors:

```powershell
Hostname = $Request.Host.ToLowerInvariant()
Scheme = $Request.Scheme.ToLowerInvariant()
```

Badgerati (changes requested):

src/Private/Logging.ps1
chris-dma: @Badgerati thank you, I have made this change now.

src/Private/Logging.ps1
chris-dma: That is a good idea and makes more sense.

src/Private/Logging.ps1
chris-dma: That is great. Much better than having a separate logic block.

Badgerati (approved):

PR #1496, opened 2025-02-21
Ensure Consistent Enum Key Usage for Pode Events

mdaneri:
This PR fixes an issue where `Pode.PodeServerEventType` was being used directly as a hashtable key in `$PodeContext.Server.Events`. Since PowerShell hashtable keys are expected to be strings, this inconsistency could cause lookup failures when retrieving or managing events.  

### **Key Highlights:**  
- **Ensured consistency** by converting `Pode.PodeServerEventType` values to strings (`$Type.ToString()`) before using them as keys.  
- **Fixed potential issues** when unregistering, testing, or retrieving events due to enum-object mismatch.  
- **Updated the following functions:**  
  - `Register-PodeEvent`  
  - `Unregister-PodeEvent`  
  - `Test-PodeEvent`  
  - `Get-PodeEvent`  
  - `Clear-PodeEvent`  

 
```powershell
# Example of the fix in action:
Register-PodeEvent -Type Start -Name 'Event1' -ScriptBlock { }
Unregister-PodeEvent -Type Start -Name 'Event1'  # Now works consistently
```
 
 Fix: https://github.com/Badgerati/Pode/issues/1497

### DIFF
--- a/src/Private/Events.ps1
+++ b/src/Private/Events.ps1
@@ -6,12 +6,12 @@ function Invoke-PodeEvent {
     )
 
     # do nothing if no events
-    if (($null -eq $PodeContext.Server.Events) -or ($PodeContext.Server.Events[$Type].Count -eq 0)) {
+    if (($null -eq $PodeContext.Server.Events) -or ($PodeContext.Server.Events[$Type.ToString()].Count -eq 0)) {
         return
     }
 
     # invoke each event's scriptblock
-    foreach ($evt in $PodeContext.Server.Events[$Type].Values) {
+    foreach ($evt in $PodeContext.Server.Events[$Type.ToString()].Values) {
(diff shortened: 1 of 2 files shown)


Badgerati:
@mdaneri well I guess this needs to be in 2.12! 😂 I'm happy with what's in the commit, and can merge if nothing else needed.

When this is merged, I'm going to prep 2.12 for release. The others will have to be 2.13 - I'm close to finishing something for the serialization PR, but it got bigger than I expected!

mdaneri:
this is bug so has to be in 2.12 
I had hope that service was going in 2.12
anyway there is also this fix to include https://github.com/Badgerati/Pode/pull/1460

Badgerati (approved):

PR #1507, opened 2025-03-01
Fixes the way logs are removed when using MaxDays, and servers restart daily; Fixes Pode removing more log files than expected due to too broad of a filter

nightroman:
This PR fixes two issues described below.

Given this file logger

```powershell
	$logging = New-PodeLoggingMethod -File -Path C:\TEMP -Name pode -MaxDays 1
	Enable-PodeRequestLogging $logging
	Enable-PodeErrorLogging $logging
```

**Issue 1: `-MaxDays` may not work**

Pode does not remove old files if the server restarts daily or more frequently (I think this is a typical local scenario).

Why. On starting `NextClearDown` is set to `[DateTime]::Now.Date` and then it is checked as `NextClearDown -lt [DateTime]::Now.Date`. This condition is always false on daily restarts and removal of old logs never happens.

**Issue 2: Pode may remove wrong logs**

This original code is supposed to remove all "old" `*.log` files in `C:\TEMP`, not necessarily related to the specific file logger "pode"

```powershell
	$null = Get-ChildItem -Path $options.Path -Filter '*.log' -Force |
		Where-Object { $_.CreationTime -lt $date } |
		Remove-Item -Force
```

I think the filter should be `"$($options.Name)_*.log"` in order to remove relevant files only.

### DIFF
--- a/src/Private/Logging.ps1
+++ b/src/Private/Logging.ps1
@@ -63,10 +63,10 @@ function Get-PodeLoggingFileMethod {
         $item.ToString() | Out-File -FilePath $path -Encoding utf8 -Append -Force
 
         # if set, remove log files beyond days set (ensure this is only run once a day)
-        if (($options.MaxDays -gt 0) -and ($options.NextClearDown -lt [DateTime]::Now.Date)) {
+        if (($options.MaxDays -gt 0) -and ($options.NextClearDown -le [DateTime]::Now.Date)) {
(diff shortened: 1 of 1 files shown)


nightroman:
Any feedback? Accepted or not? When to expect the fix released, approximately?

Badgerati:
Yep, changes all look good. Sorry, just very busy with work atm.

I'm going to look at making a 2.12.1 release, though can't promise on when it will be released, but I'll aim for in the next couple weeks.

PR #1509, opened 2025-03-01
Support for Custom and default Favicons in Pode Endpoints

mdaneri:
#### **Overview**

This PR introduces a dedicated set of favicon management functions, replacing the previous `-Favicon` and `-DefaultFavicon` parameters on `Add-PodeEndpoint`. Favicons can now be configured more flexibly and consistently across one or more endpoints, after endpoints have been created.

---

#### **Changes**

- **Added:**  
  - `Add-PodeFavicon`: Assigns a custom or default favicon to one or more HTTP/S endpoints  
  - `Remove-PodeFavicon`: Removes the favicon from one or more endpoints  
  - `Get-PodeFavicon`: Retrieves favicon data from configured endpoints  
  - `Test-PodeFavicon`: Checks if favicons are configured on endpoints

- All functions support `-EndpointName` and `-DefaultEndpoint` targeting
- New documentation created for favicon management

---

#### **Usage Examples**

```powershell
# Create an HTTP endpoint
Add-PodeEndpoint -Address localhost -Port 8080 -Protocol Http

# Add default Pode favicon to all endpoints
Add-PodeFavicon -Default

# Add custom favicon to only default endpoints
$iconBytes = [System.IO.File]::ReadAllBytes("C:\path\to\custom.ico")
Add-PodeFavicon -Binary $iconBytes -DefaultEndpoint

# Remove all favicons
Remove-PodeFavicon

# Check if default endpoints have a favicon set
Test-PodeFavicon -DefaultEndpoint
```

---

For favicon format guidance and recommendations, refer to:
- [[Favicon specification]](https://en.wikipedia.org/wiki/Favicon)  
- [[RFC 5988]](https://datatracker.ietf.org/doc/html/rfc5988)


Badgerati (changes requested):

src/Public/Endpoint.ps1
```diff
@@ -292,6 +303,30 @@ function Add-PodeEndpoint {
...
+
+        # Check if running in PowerShell Core to determine the correct method for reading binary data.
+        if (Test-PodeIsPSCore) {
+            # In PowerShell Core, use -AsByteStream to read the file as a byte array.
+            $Favicon = (Get-Content -Path ([System.IO.Path]::Combine($podeRoot, 'favicon.ico')) -Raw -AsByteStream)
+        }
+        else {
+            # In Windows PowerShell, use -Encoding byte to read the file as a byte array.
+            $Favicon = (Get-Content -Path ([System.IO.Path]::Combine($podeRoot, 'favicon.ico')) -Raw -Encoding byte)
+        }
```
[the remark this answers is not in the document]
mdaneri: there are 2 places where a this logic is used. 
Here and `Write-PodeFileResponseInternal` I'm not sure is worth it

src/Private/PodeServer.ps1
```diff
@@ -210,59 +210,66 @@ function Start-PodeWebServer {
...
-                                        Group       = $WebEvent.Request.SseClientGroup
-                                        ClientId    = $WebEvent.Request.SseClientId
-                                        LastEventId = $null
-                                        IsLocal     = $false
-                                    }
+                                
+                                # deal with favicon if available
+                                if ($WebEvent.Path -eq '/favicon.ico' -and ($null -ne $PodeContext.Server.Endpoints[$context.EndpointName].Favicon)) {
+                                    # Write the file content as the HTTP response
+                                    Write-PodeTextResponse -Bytes $PodeContext.Server.Endpoints[$context.EndpointName].Favicon -ContentType 'image/png' -StatusCode 200
```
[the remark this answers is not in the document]
mdaneri: I create a function to detect the image type

src/Public/Endpoint.ps1
```diff
@@ -292,6 +303,30 @@ function Add-PodeEndpoint {
...
 
+    # Check if both -Favicon and -DefaultFavicon are provided, which is not allowed.
+    # If both are set, throw an exception with a relevant message.
+    if (($null -ne $Favicon) -and $DefaultFavicon) {
+        throw ($Podelocale.parametersMutuallyExclusiveExceptionMessage -f '-Favicon', '-DefaultFavicon')
+    }
+
+    # If no -Favicon is provided, the protocol is either HTTP or HTTPS, and -DefaultFavicon is enabled,
+    # set the default favicon from the Pode module's miscellaneous path.
+    if ( ($null -eq $Favicon) -and (@('Http', 'Https') -icontains $Protocol) -and $DefaultFavicon) {
```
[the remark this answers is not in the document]
mdaneri: Why? I can call add-podeEndpoint with both -Favicon adn -DefaultFavicon.
I can create an new parameterset but start to become complicated with all these parameters

src/Public/Endpoint.ps1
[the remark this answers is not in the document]
mdaneri: I solved using [System.IO.File]::ReadAllBytes($Favicon) it's much better and works everywhere

Badgerati:
I was having a think about this yesterday, and I'm wondering if it's actually better to separate the Favicon logic away from Endpoint. Separating it out into it's own `Add-PodeFavicon` will allow use to support both differing favicons for Endpoints, but also for Routes as well - as we could have an `-EndpointName` and `-Route` parameters, where `-Route` is either a literal path or path pattern much like on `Add-PodeMiddleware`.

This helps keep the Endpoint functions tidier, but also lets us more easily extend Favicon support in the future if we need to.

As for where the favicon selector logic runs, currently is directly in the PodeServer.ps1 scriptblock. I'm thinking this would be better either in the `Get-PodePublicMiddleware` (at the top, before we check `/public`) - this way all supported web servers get it. Or, we add a new `Get-PodeFaviconMiddleware` which get run before `Get-PodePublicMiddleware`.

This would also mean you could have different `-Path`, `-Binary`, `-Default` parameters split by a ParameterSet - so it's more obvious/natural as to what the parameter does.

Ie:
```powershell
# load file and read in bytes
Add-PodeFavicon
    -Path [string]
    -EndpointName [string]
    -Route [string]

# has bytes directly supplied for custom cases
Add-PodeFavicon
    -Binary [byte[]]
    -EndpointName [string]
    -Route [string]

# use the internal default favicon
Add-PodeFavicon
    -Default [switch]
    -EndpointName [string]
    -Route [string]
```

EndpointName/Route could be `[string[]]` if we want. They are also optional, so if `Add-PodeFavicon -Default` is only supplied, this applies to every request.

Additional functions:

* `Remove-PodeFavicon`
* `Test-PodeFavicon`
* `Get-PodeFavicon`

In terms of the fitlering/priority for selecting the favicon, if the request is `/favicon.ico`:

1. If no custom favicons, skip and let `/public` try and use `favicon.ico`
2. If custom favicons:
    a. Filter by EndpointName
    b. Filter by Route on the above EndpointName filtering - if the above returned an empty array then filter on the original array for favicons with no EndpointName
    c. If still no favicons, then get all favicons from the original array with no EndpointName or Route (ie, a favicon for every request)
    c. If favicon found, return custom/default favicon
    e. In the case where multiple favicons match, either choose the last one (ie, the last favicon created), or possibly a `-Priority` flag on `Add-PodeFavicon` 🤔

mdaneri:
I was rethinking your proposal
I agree regarding
Remove-PodeFavicon
Test-PodeFavicon
Get-PodeFavicon

I'm going to implement them

Badgerati:
I don't think it's that overkill, where it is in PodeServer.ps1 currently is basically middleware, but I agree that the `-Route` param could be dropped here and we just have `-EndpointName`.

I've read though RFC5988 and RFC8288, and I can't see anything that says "the favicon should be returned regardless of any security measures or routing rules", nor anything in any other RFC or the HTML5 docs - so you're going to need to link this to me, from what I've read servers can return a 401, 403, and 429 for the favicon, which suggests servers can apply security measures. Given what I'm reading, I see no issue in this being a part of the /public inbuilt middleware.

mdaneri:
sorry for the delay. I had a very busy week

I implemented the new cmdlets

Badgerati:
> sorry for the delay. I had a very busy week
> 
> I implemented the new cmdlets

Don't worry about it, I've had a very busy month! 😄 

---

On the Favicon in middleware and requests a page you don't have access to/etc., would that not be resolved by placing making the favicon be the very first in the list of inbuilt middlewares?

Ie, add a new `Get-PodeFaviconMiddleware` in `Private/Middleware.ps1` and then in PodeServer.ps1 and Serverless.ps1 update the the inbuilt middleware lists?

```powershell
    $inbuilt_middleware = @(
        (Get-PodeFaviconMiddleware),
        (Get-PodeSecurityMiddleware),
        (Get-PodeAccessMiddleware),
        (Get-PodeLimitMiddleware),
        (Get-PodePublicMiddleware),
        (Get-PodeRouteValidateMiddleware),
        (Get-PodeBodyMiddleware),
        (Get-PodeQueryMiddleware),
        (Get-PodeCookieMiddleware)
    )
```

^ that would be like how you have it hardcoded in PodeServer.ps1 currently - but also accessible for Serverless.ps1

src/Private/Helpers.ps1
```diff
@@ -3986,3 +3986,65 @@ function ConvertTo-PodeSleep {
...
+
+.EXAMPLE
+    $bytes = [System.IO.File]::ReadAllBytes("C:\path\to\image.gif")
+    $mimeType = Get-PodeImageContentType -Image $bytes
+    # $mimeType will be 'image/gif' if the image is a GIF.
+    
+.NOTES
+    This is an internal function and may change in future releases of Pode.
+#>
+function Get-PodeImageContentType {
```
Badgerati, line 4014: Could we change this function to do direct array indexing instead, for checking the bytes? Remove the array cloning, and string checking; direct byte checking would be more efficient and faster.

I've also updated the `<svg` check to be a byte check as well:

```powershell
function Get-PodeImageContentType {
    param(
        [Parameter()]
        [byte[]]
        $Image
    )

    if (($null -eq $Image) -or ($Image.Length -lt 12)) {
        return 'application/octet-stream'
    }

    # WebP: starts with "RIFF....WEBP"
    if (
        $Image[0] -eq 0x52 -and $Image[1] -eq 0x49 -and $Image[2] -eq 0x46 -and $Image[3] -eq 0x46 -and
        $Image[8] -eq 0x57 -and $Image[9] -eq 0x45 -and $Image[10] -eq 0x42 -and $Image[11] -eq 0x50
    ) {
        return 'image/webp'
    }

    # PNG: starts with "89 50 4E 47 0D 0A 1A 0A"
    if (
        $Image[0] -eq 0x89 -and $Image[1] -eq 0x50 -and $Image[2] -eq 0x4E -and $Image[3] -eq 0x47 -and
        $Image[4] -eq 0x0D -and $Image[5] -eq 0x0A -and $Image[6] -eq 0x1A -and $Image[7] -eq 0x0A
    ) {
        return 'image/png'
    }

    # GIF (GIF87a or GIF89a): starts with "47 49 46 38 37 61" or "47 49 46 38 39 61"
    if (
        $Image[0] -eq 0x47 -and $Image[1] -eq 0x49 -and $Image[2] -eq 0x46 -and $Image[3] -eq 0x38 -and
        (
            ($Image[4] -eq 0x37 -and $Image[5] -eq 0x61) -or
            ($Image[4] -eq 0x39 -and $Image[5] -eq 0x61)
        )
    ) {
        return 'image/gif'
    }

    # ICO: starts with "00 00 01 00"
    if (
        $Image[0] -eq 0x00 -and $Image[1] -eq 0x00 -and
        $Image[2] -eq 0x01 -and $Image[3] -eq 0x00
    ) {
        return 'image/x-icon'
    }

    # SVG: starts with "<svg"
    if (
        $Image[0] -eq 0x3C -and $Image[1] -eq 0x73 -and
        $Image[2] -eq 0x76 -and $Image[3] -eq 0x67
    ) {
        return 'image/svg+xml'
    }

    # JPEG: starts with "FF D8 FF"
    if (
        $Image[0] -eq 0xFF -and $Image[1] -eq 0xD8 -and $Image[2] -eq 0xFF
    ) {
        return 'image/jpeg'
    }

    # If none of the above formats match, return a generic binary type
    return 'application/octet-stream'
}
```

src/Public/FavIcon.ps1
```diff
@@ -0,0 +1,346 @@
...
+	the favicon to all HTTP/S endpoints, a specific endpoint, or only to endpoints marked as default.
+
+.PARAMETER Default
+	Use the default Pode favicon.ico embedded in the module.
+
+.PARAMETER Path
+	The path to a custom favicon.ico file to use. Must be a valid, accessible path.
+
+.PARAMETER Binary
+	A raw byte array representing the favicon.ico file contents.
```
Badgerati, line 17: Could we tweak the reference to a `favicon.ico file` in the Path/Binary parameter descriptions to just be to a `favicon file`, and specify the possible image types? Just to avoid confusion that it can be a png or svg as well - it doesn't *need* to be an ico file

src/Public/FavIcon.ps1
```diff
@@ -0,0 +1,346 @@
...
+
+.PARAMETER EndpointName
+	The name of the specific endpoint to apply the favicon to. If not provided, the favicon is applied to all HTTP/S endpoints.
+
+.PARAMETER DefaultEndpoint
+	If supplied, only endpoints marked with `.Default = $true` will receive the favicon.
```
Badgerati, line 23: For each of the `DefaultEndpoint` parameter descriptions, I would start it with the following instead:

```
If supplied, only endpoints created with `-Default` ...
```

As it's more what the user will see in their code, rather than the internal `.Default = $true`

src/Public/FavIcon.ps1
```diff
@@ -0,0 +1,346 @@
...
+            continue
+        }
+
+        # If any matched endpoint lacks a favicon, return false
+        if ($null -eq $endpoint.Favicon) {
+            return $false
+        }
+    }
+
+    return $true
```
Badgerati, line 200: Lines 186-200 for the `foreach` and `return` will want tweaking to the following - so we skip non-HTTP endpoints and return false more appropriately when no default/HTTP endpoints are found:

```powershell
$found = $false
foreach ($key in $keys) {
    $endpoint = $PodeContext.Server.Endpoints[$key]

    # Only check HTTP/S endpoints
    if (@('Http', 'Https') -inotcontains $endpoint.Protocol) {
        continue
    }

    # If filtering to default endpoints only, skip others
    if ($DefaultEndpoint -and !$endpoint.Default) {
        continue
    }

    # If the endpoint has no favicon, return false
    if ($null -eq $endpoint.Favicon) {
        return $false
    }

    # If we reach here, the endpoint has a favicon
    $found = $true
}

return $found
```

src/Private/PodeServer.ps1
```diff
@@ -8,6 +8,7 @@ function Start-PodeWebServer {
 
     # setup any inbuilt middleware
     $inbuilt_middleware = @(
```
mdaneri, line 10: Limit should be the first

src/Public/FavIcon.ps1
```diff
@@ -0,0 +1,346 @@
+<#
+.SYNOPSIS
+	Adds a favicon to one or more Pode HTTP/HTTPS endpoints.
```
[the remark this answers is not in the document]
mdaneri: I don't know why this happened.

I have to create something to fix all these comments

src/Public/FavIcon.ps1
mdaneri: you are right

src/Public/FavIcon.ps1
```diff
@@ -0,0 +1,346 @@
...
+    foreach ($key in $keys) {
+        $endpoint = $PodeContext.Server.Endpoints[$key]
+
+        # If filtering to default endpoints only, skip others
+        if ($DefaultEndpoint -and !$endpoint.Default) {
+            continue
+        }
+
+        # Only include endpoints that have a favicon set
+        if ($null -ne $endpoint.Favicon) {
```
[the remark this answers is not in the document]
Badgerati: Yeah, the `if` check on line 270; so:

```powershell
```

src/Private/PodeServer.ps1
Badgerati: Yep, sounds good to me

Badgerati (approved):

Issue #1500, opened 2025-02-23
`Write-PodeFileResponse` returns 404 when a path contains square brackets


helpimnotdrowning:
### Describe the Bug
I am learning Pode by writing a file browser. When trying to serve a file whose name contains square brackets (`/mnt/URUHA/mnt/W/ytdl/Matsuri Channel 夏色まつり/20230115 - 【歌枠】新エフェクターでケロケロ歌う！_Singing KARAOKE【ホロライブ_夏色まつり】/Matsuri Channel 夏色まつり - 20230115 - 【歌枠】新エフェクターでケロケロ歌う！_Singing KARAOKE【ホロライブ_夏色まつり】 [n82XjwgrQ54].jpg`), Pode returns 404.

I created a file next to that one with the square brackets removed (`.../Matsuri Channel 夏色まつり - 20230115 - 【歌枠】新エフェクターでケロケロ歌う！_Singing KARAOKE【ホロライブ_夏色まつり】 n82XjwgrQ54.jpg`), and this file downloads normally.

Manually escaping the filename
```pwsh
$path -replace '\[','`[' -replace '\]', '`]'
```
functions as a workaround.

Pode seems to resolve wildcard characters in file paths.

### Steps To Reproduce
1. Create a file "[hello].txt" with contents "brackets file"
2. Create a file "hi.txt" with contents "no brackets file"
3. Run the following Pode server in the same directory:
```pwsh
import-module pode
Start-PodeServer {
	Add-PodeEndpoint -Address localhost -Port 8080 -Protocol Http
	
	Add-PodeRoute -Method GET -Path '/*' -ScriptBlock {
		$path = "$pwd$($Webevent.path)"
		
		if (Test-Path -type leaf -literalpath $path) {
			Write-PodeFileResponse -path $path
		} else  {
			Write-PodeHtmlResponse -Value @"
<a href="[hello].txt">brackets</a>
<a href="hello.txt">no brackets</a>
"@
		}
	}
}
```
4. Click on the "brackets" link. It should show Pode's 404 error page.
5. Click on the "no brackets" link. It should show the contents of the "hi.txt" file, "no brackets file"
6. Replace the line `Write-PodeFileResponse ...` with
```pwsh
Write-PodeFileResponse -path ($path -replace '\[','`[' -replace '\]', '`]')
```
7. The "brackets" file should now be viewable.

### Expected Behavior
[...]

PR #1522, opened 2025-03-15
Enable Configurable CDN for Default Pode Static Pages

mdaneri:
## **Summary**
This pull request introduces a new feature that allows users to configure the **CDN source** for serving default Pode static pages, such as **Swagger UI**. This improves reliability and performance by allowing users to choose from multiple CDN providers.

## **Why This Change?**
Currently, Pode defaults to using `https://unpkg.com` for serving static assets. However, **unpkg has proven to be unreliable** due to the following reasons:
- **Frequent Downtime**: `unpkg.com` has experienced periods of unavailability, causing failures in serving assets.  
- **No Built-in Redundancy**: Unlike `jsDelivr`, `unpkg.com` lacks automatic fallback mechanisms.  
- **Performance Issues**: Users have reported inconsistent load times affecting page rendering.  
- **Enterprise Restrictions**: Some corporate environments block `unpkg.com`, leading to accessibility issues.  

To address these concerns, this change introduces a **configurable CDN option**, allowing users to select a more stable and performant source.

## **Feature Details**
A new `ContentDeliveryNetwork` setting is introduced under `Web.Static` in `server.psd1`. This setting allows users to specify which CDN should be used for serving default static assets.

### **Supported CDNs**
- `https://cdn.jsdelivr.net/npm` *(default)*
- `https://unpkg.com`
- `https://cdnjs.cloudflare.com/ajax/libs` *(requires URL modifications)*

## **Example Configuration in `server.psd1`**
Users can specify their preferred CDN in their Pode configuration:

```powershell
@{
    RestFulPort = 8081
    Protocol = 'Http'
    Address = 'localhost'
    Server = @{
        Timeout = 60
        BodySize = 100MB
    }
    Web = @{
        OpenApi = @{
            DefaultDefinitionTag = 'v3.0.3'
        }
        Static = @{
            ContentDeliveryNetwork = 'https://cdn.jsdelivr.net/npm'
        }
    }
}
```

## **Implementation Notes**
- The default CDN is `https://cdn.jsdelivr.net/npm` to provide better reliability.
- Users can override this setting by specifying a different CDN in their configuration.
- If the `ContentDeliveryNetwork` key is missing, Pode will fall back to the default (`jsDelivr`).
- The change ensures backward compatibility with existing Pode configurations.

## **Testing**
- Verified that Pode correctly loads Swagger UI from different CDN sources.
- Tested fallback behavior when no CDN is specified.
- Ensured no breaking changes for users who do not set this configuration.

## **Benefits of This Change**
✔ **Increased Reliability** – Users can avoid any CDN that has downtime by switching to a more stable CDN.  
✔ **Flexibility** – Users can choose the best CDN for their environment, avoiding corporate restrictions.  
✔ **Seamless Integration** – Backward compatible with existing configurations.
 
Fix: #1523


Badgerati:
Would it be worth, similar to Pode.Web, if we bundled the packages with Pode so it can use them internally, and allow people with air-happed networks to use Swagger?

If we do, we can probably use Yarn again - inline with Pode.Web. And we'd probably need to setup an internal static route, so swagger/etc can get to the JS/CSS.

mdaneri:
Sure, we can do that. However, my understanding was that you didn’t want it. If that’s the case, I’ll leave it to you since you already did with pote.web.

Badgerati:
In the OpenAPI work it was mostly because existing pages already used libraries from a CDN, so it was just conforming to those - plus unlike in Pode.Web, it hadn't been asked about air-gapped networks.

But if we need to change now because unpkg is unreliable, and it's now been asked in Discord about air-gapped networks for Swagger, we might as well.

If you want me to do it like with Pode.Web, I can :) I just need to finish off a fix for #1500

Badgerati:
Closing, as superseded by #1538

Issue #1527, opened 2025-03-27
Web socket regression on 2.12.0


fatherofinvention:
### Describe the Bug
Web sockets don't work in Pode 2.12.0 but do work in 2.11.1. 

### Steps To Reproduce
Just setup a ws endpoint and have a client send messages to it. You will see error in the terminal:

```
[Error] ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at System.MemoryExtensions.AsMemory[T](T[] array, Int32 start, Int32 length)
   at Pode.PodeRequest.Receive(CancellationToken cancellationToken) in C:\Projects\pode-builds\2.12.0\src\Listener\PodeRequest.cs:line 264
```

SERVER:
```
Add-PodeSignalRoute -Path '/ws' -ScriptBlock { 
  # logic here
}
```

CLIENT:
```
  <html>
  <head>
  <title>Websocket Test</title>
  <script type="text/javascript">
  var ws = new WebSocket("ws://localhost:8080/ws");
  ws.onopen = function() {
    document.getElementById('status').innerHTML = "Connected";
  };
  ws.onmessage = function (evt) {
    document.getElementById('results').innerHTML += "<p>" + evt.data + "</p>";
    // Send acknowledgement back to server when a message is received
    ws.send(JSON.stringify({ Message: 'hello from client' }));
  };
  ws.onclose = function() {
    document.getElementById('status').innerHTML = "Closed";
  };
  </script>
  </head>
  <body>
  <h1>Websocket Test</h1>
  <div>Status: <span id="status">Connecting...</span></div>
  <div id="results"></div>
  </body>
  </html>

```

### Platform
 - OS: Ubuntu 22.04.5 LTS
 - Versions:
   - Pode: 2.12.0
   - PowerShell: 7.5.0

PR #1532, opened 2025-03-30
Fix buffer reuse issue in WebSockets

mdaneri:
**Description:**  
This pull request addresses issue [[#1527]](https://github.com/Badgerati/Pode/issues/1527)  by modifying the way buffers are managed in PodeSignalRequest. The problem was that the base class, PodeRequest, uses a readonly, preallocated buffer that cannot be reinitialized. This caused issues for WebSocket signals (handled by PodeSignalRequest) where a fresh buffer is needed per Receive call.

**Changes Introduced:**

1. **Lazy Buffer Initialization in PodeRequest:**  
   - Removed the immediate allocation of the buffer from the constructor.
   - Introduced lazy initialization of the buffer using a private backing field with a property:
     ```csharp
     private byte[] _buffer;
     protected virtual byte[] Buffer
     {
         get
         {
             if (_buffer == null)
             {
                 _buffer = new byte[BufferSize];
             }
             return _buffer;
         }
     }
     ```
   - Added a virtual method `GetBuffer()` that returns the buffer. By default, it returns the lazily initialized, single preallocated buffer:
     ```csharp
     protected virtual byte[] GetBuffer() => Buffer;
     ```

2. **Override in PodeSignalRequest:**  
   - In PodeSignalRequest, override `GetBuffer()` to return a new byte array for each Receive call:
     ```csharp
     protected override byte[] GetBuffer() => new byte[BufferSize];
     ```
   This ensures that every WebSocket signal request gets a fresh buffer, avoiding any leftover data from previous operations.

3. **Integration Test:**  
   - Added a Pester integration test that creates a WebSocket client to send a JSON signal and asserts that the returned message (e.g. a date string) is correctly received. This test mimics the client behavior from our JS code and ensures that the buffer management in PodeSignalRequest is working correctly.

**Impact:**  
- These changes improve memory management by only allocating the shared buffer when necessary for most requests.
- For WebSocket signals, a new buffer is used on each receive call, ensuring data integrity.
- The modifications are backward-compatible for other request types (e.g., HTTP requests) that benefit from reusing the preallocated buffer.

**Testing:**  
- Verified the changes locally using the provided Pester integration test.
- The WebSocket client test successfully sends a signal and receives the correct response without any buffer contamination issues.

### DIFF
--- a/src/Listener/PodeRequest.cs
+++ b/src/Listener/PodeRequest.cs
@@ -231,8 +231,9 @@ namespace Pode
         /// <summary>
         /// Provides access to a buffer. The buffer is allocated only when first requested,
         /// saving memory if it is never needed.
+        /// This property is virtual to allow derived classes to override the buffer allocation behavior.
         /// </summary>
-        protected byte[] Buffer
+        protected virtual byte[] Buffer
         {
             get
             {
@@ -243,13 +244,6 @@ namespace Pode
                 return _buffer;
             }
         }
-
-        /// <summary>
-        /// This virtual method returns a buffer to be used during the receive operation.
-        /// Derived classes can override this to customize buffer allocation behavior.
-        /// By default, it returns the lazily allocated Buffer property.
-        /// </summary>
-        protected virtual byte[] GetBuffer() => Buffer;
 
         /// <summary>
         /// Receives data from the input stream and processes it.
@@ -266,7 +260,7 @@ namespace Pode
                 }
 
                 Error = null;
-                var localBuffer = GetBuffer();
+                var localBuffer = Buffer;
                 using (BufferStream = new MemoryStream())
                 {
                     var close = true;
@@ -364,7 +358,7 @@ namespace Pode
             }
 
             // Read data from the input stream until the check bytes are found
-             var localBuffer = GetBuffer();
+             var localBuffer = Buffer;
             using (var bufferStream = new MemoryStream())
             {
                 while (true)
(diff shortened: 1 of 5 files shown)


src/Listener/PodeRequest.cs
```diff
@@ -229,6 +228,29 @@ private bool ValidateCertificateCallback(object sender, X509Certificate certific
...
+                return _buffer;
+            }
+        }
+
+        /// <summary>
+        /// This virtual method returns a buffer to be used during the receive operation.
+        /// Derived classes can override this to customize buffer allocation behavior.
+        /// By default, it returns the lazily allocated Buffer property.
+        /// </summary>
+        protected virtual byte[] GetBuffer() => Buffer;
```
Badgerati, line 252: Any specific reason to have this method, and not just override the `Buffer` property in PodeSignalRequest with the following instead?

```c#
protected override byte[] Buffer
{
    get => new byte[BufferSize];
}
```

Badgerati: But the `Buffer` property is marked as `virtual` 👀

Badgerati: This was my main question, why use the method and not the property? There shouldn't be any issues, and I've done similar in other places with virtual properties; the method feels a tad redundant 🤔

mdaneri:
Added a note to the documentation regarding the current limit of 32KB payload

Badgerati (approved):
LGTM

PR #1537, opened 2025-04-13
Adds support for fragmented WebSocket frames, and larger payloads

Badgerati:
### Description of the Change
Fixes a limit found in #1532 around websockets, where payloads could only be up to 32KB (PowerShell version depending).

The fixes applied are:

* Adds support for fragmented frame on request and response
* Adds support to wait for more data on larger request payloads


mdaneri:
Can you uncomment the last integration test for websocket?

Badgerati:
@mdaneri it is uncommented 👀

mdaneri:
My bad🙂

TheBSGamer:
Seems to be working perfectly aside from some notes, not sure if they're related. Tested with a 18MB file and it sent the socket as expected.

I did notice in the logs that I am getting a timeout error on the verbose logs still:
```
[Verbose]: [ContextId: aef3c2e0-f43e-cb95-9c68-b98fa5d3e673] Response timed-out sent
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at System.Threading.CancellationToken.ThrowIfCancellationRequested()
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource<System.Int32>.GetResult(Int16 token)
   at Pode.PodeRequest.Receive(CancellationToken cancellationToken) in C:\Users\Username\Desktop\Scripts\Pode\src\Listener\PodeRequest.cs:line 281
[Verbose]: [ContextId: aef3c2e0-f43e-cb95-9c68-b98fa5d3e673] Response disposed
```

Everything is working as expected though. The error does seem to log a good five-ish seconds after the response is handled, so I'm not really sure what the issue is. I can provide the test repo if needed. Just created a root view page, a web socket that responded with a success message, and a (regular) route to broadcast a large payload over the socket. Initiated the connection through the console on the root view page.

Badgerati:
Hi @TheBSGamer,

Those verbose timeout messages are fine, the initial HTTP connection made to initialise the WebSocket is what will be timing out.

That being said, I've just pushed a new commit that reduces unneeded timeout messages - such as the one you're seeing; the request was being upgraded from HTTP to WebSocket, but the timeout callback was being disposed.

If you try now, you should see reduced timeout messages from WebSockets.

Issue #1545, opened 2025-04-23
ConvertTo-PodeYamlInternal fails to quote strings starting with YAML-special characters


mdaneri:
**Description:**  
When a value begins with one of YAML’s special prefix characters (`-`, `?`, `:`, `,`, `[`, `]`, `{`, `}`, `#`, `&`, `*`, `!`, `|`, `>`, `'`, `"`, `%`, `@`, `` ` ``), our converter emits it unquoted. That can lead to invalid or mis-parsed YAML:

```powershell
PS> ConvertTo-PodeYamlInternal -InputObject '-dash' -NoNewLine
-dash
```

YAML will treat this as a list item or syntax error, rather than the literal string `"-dash"`.

**Steps to Reproduce:**  
1. Call on any string beginning with a special character:
   ```powershell
   ConvertTo-PodeYamlInternal -InputObject '-example' -NoNewLine
   ```
2. Notice the output is unquoted.

**Actual Behavior:**  
Special-prefixed values are never wrapped in quotes, causing them to be parsed incorrectly.

**Expected Behavior:**  
Any string whose first character is a YAML-special prefix must be quoted:
```yaml
'-dash'
```

Issue #1544, opened 2025-04-23
`ConvertTo-PodeYamlInternal` emits platform newlines (CRLF) and incorrect indentation for stand-alone block literals


mdaneri:
**Description:**  
The YAML converter currently uses `[Environment]::NewLine` when building multi-line (“|”) scalars and applies extra spaces assuming a nested context. On Windows this produces `\r\n` line endings, and tests (and many YAML consumers) expect LF-only (`\n`). It also indents stand-alone block literals by two spaces unnecessarily.

---

**Steps to Reproduce:**  
1. Call the converter on a simple multi-line string in PowerShell on Windows:  
   ```powershell
   ConvertTo-PodeYamlInternal -InputObject "line1`nline2" -NoNewLine

PR #1546, opened 2025-04-23
Fixes unquoted YAML strings starting with special characters, incorrect indentation for standalone block literals, and incorrect line endings (CRLF)

mdaneri:
**Commit Message:**  
Quote strings that begin with YAML-special prefix characters as single-quoted scalars to ensure valid YAML output.

**Description:**  
Ensure that any scalar starting with a YAML-special prefix character is emitted as a single-quoted string.

**Additional Fix:**  
The YAML converter currently uses `[Environment]::NewLine` when building multi-line (`"|"`) scalars and applies extra spaces assuming a nested context. On Windows this produces `
` line endings, and tests (and many YAML consumers) expect LF-only (`
`). It also indents stand-alone block literals by two spaces unnecessarily.

**Changes:**  
```diff
--- src/Public/ConvertTo-PodeYamlInternal.ps1
+++ src/Public/ConvertTo-PodeYamlInternal.ps1
@@ <string> case before returning unquoted literal:
-                else {
-                    if ($string -match '^[#\[\]@\{\}\!\*]') {
-                        "'$($string -replace '''', '''''')'"
-                    }
-                    else {
-                        $string
-                    }
+                else {
+                    # detect YAML-special prefix and force quoting
+                    $isSpecialPrefix = $string -match '^[\-\?:,\[\]{}#&*!|>''"%@`]'
+                    if ($isSpecialPrefix) {
+                        # single-quote style: escape internal ' by doubling
+                        $escaped = $string -replace "'", "''"
+                        "'$escaped'"
+                    }
+                    else {
+                        $string
+                    }
                 }
```

**Verification Steps:**  
1. Run `ConvertTo-PodeYamlInternal -InputObject '-dash' -NoNewLine` → returns `'-dash'`.  
2. `Invoke-Pester` → “special prefix” test passes.

Fix: #1545, #1544

### DIFF
--- a/src/Private/Helpers.ps1
+++ b/src/Private/Helpers.ps1
@@ -3589,12 +3589,12 @@ function ConvertTo-PodeYamlInternal {
 
                             $index += $breakpoint
                             if ($index -lt $length) {
-                                $null = $multiline.Append([System.Environment]::NewLine)
+                                $null = $multiline.Append("`n")
                             }
                         }
 
                         if ($i -lt ($items.Length - 1)) {
-                            $null = $multiline.Append([System.Environment]::NewLine)
+                            $null = $multiline.Append("`n")
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


Badgerati (approved):
LGTM if you're happy for me to merge?

mdaneri:
Yes

PR #1634, opened 2025-11-17
fix net10 build: comparison with string and int that worked until now

helpimnotdrowning:
### Description of the Change
Building on net10 failed because the build script tried comparing the major version `"10"` (a string) with `9` (an integer). I haven't checked how string/int comparisons work exactly, but `"9" -ge 9` is `True`, but `"10" -ge 9` is `False`. The script would wrongly find that net10 isn't able to build for _any_ version of the framework and fail.

This patch just converts it to an integer.

Using:
- Debian 13, amd64
- dotnet 10.0.100 (via MS apt repo)


Badgerati:
Spotted this issue myself while working on #1638 and fixed it there, I did it slightly differently though as I simplified the logic to just use `dotnet --version` - that's basically what the original logic boiled down to anyway.

PR #1710, opened 2026-05-19
Fix for default folders whose custom path is not recognized

### LINKED ISSUE #1708: Unable to define a custom path for Pode default folders
### Describe the Bug
I want to customize the path to the folder containing the views dedicated to errors (cf. https://badgerati.github.io/Pode/Tutorials/Routes/Utilities/ErrorPages/).

To do this, I use either the `Set-PodeDefaultFolder Errors <Path>` function or the `Server.DefaultFolders` option in the server configuration file.
But my custom views are never invoked. They are only invoked when I use the default `errors` folder located at the server root path (i.e. `"$(Get-PodeServerPath)/errors"`).

### Steps To Reproduce
1. In the server script block, call the `Set-PodeDefaultFolder` function with a custom path.
2. In the script block of any route, call the `Set-PodeResponseStatus` function with an error status code (e.g. `403`).
3. Open the route in a browser or a HTTP client.
4. (The default view is invoked instead of the custom one)

```powershell
# Requirement: create a "ErrorViews" folder and put a "default.html" document inside. 
Start-PodeServer {
  Add-PodeEndpoint -Address 127.0.0.1 -Port 5000 -Protocol Http
  Set-PodeDefaultFolder Errors (Join-Path (Get-PodeServerPath) "ErrorViews")
  Add-PodeRoute -Method Get -Path "/" -ScriptBlock { Set-PodeResponseStatus 403 }
}
# Fix: rename the "ErrorsViews" folder to "errors" 😢 
```

### Expected Behavior
Errors are displayed using the views contained in the folder provided as an argument to the `Set-PodeDefaultFolder` function.

### Screenshots
Actual:
Image

Expected:
Image

### Platform
 - OS: Windows 11
 - Browser: Chrome 148
 - Versions:
   - Pode: 2.13.2
   - PowerShell: 7.6.1

CedX:
Quick and dirty fix for #1708.
Probably not the right way to do it, but it gives us a starting point...
(I'm still a beginner with PowerShell, and I'm just starting to look into Pode.)

### DIFF
--- a/src/Public/Core.ps1
+++ b/src/Public/Core.ps1
@@ -948,7 +948,6 @@ function Set-PodeDefaultFolder {
     )
     if (Test-Path -Path $Path -PathType Container) {
         $PodeContext.Server.DefaultFolders[$Type] = $Path
-        $PodeContext.Server.InbuiltDrives[$Type] = (New-PodePSDrive -Path $path)
     }
     else {
         # Path does not exist
--- a/src/Public/Core.ps1
+++ b/src/Public/Core.ps1
@@ -948,6 +948,7 @@ function Set-PodeDefaultFolder {
     )
     if (Test-Path -Path $Path -PathType Container) {
         $PodeContext.Server.DefaultFolders[$Type] = $Path
+        $PodeContext.Server.InbuiltDrives[$Type] = (New-PodePSDrive -Path $path)
     }
     else {
         # Path does not exist
--- a/src/Public/Core.ps1
+++ b/src/Public/Core.ps1
@@ -946,6 +946,8 @@ function Set-PodeDefaultFolder {
         [string]
         $Path
     )
+
+    $Path = Join-PodeServerRoot -Folder $Path
     if (Test-Path -Path $Path -PathType Container) {
         $PodeContext.Server.DefaultFolders[$Type] = $Path
     }
--- a/src/Private/Helpers.ps1
+++ b/src/Private/Helpers.ps1
@@ -812,19 +812,19 @@ function Add-PodePSInbuiltDrive {
     # create drive for views, if path exists
     $path = (Join-PodeServerRoot -Folder $PodeContext.Server.DefaultFolders.Views)
     if (Test-Path $path) {
-        $PodeContext.Server.InbuiltDrives[$PodeContext.Server.DefaultFolders.Views] = (New-PodePSDrive -Path $path)
+        $PodeContext.Server.InbuiltDrives['views'] = (New-PodePSDrive -Path $path)
     }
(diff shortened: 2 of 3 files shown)


src/Public/Core.ps1
```diff
@@ -948,6 +948,7 @@ function Set-PodeDefaultFolder {
     )
     if (Test-Path -Path $Path -PathType Container) {
         $PodeContext.Server.DefaultFolders[$Type] = $Path
+        $PodeContext.Server.InbuiltDrives[$Type] = (New-PodePSDrive -Path $path)
```
Badgerati: This change isn't required, the fix for the InbuiltDrives keys is correct - they should have been `'errors'`, `'views'` and `'public'`; likely a find/replace gone rogue.

Additionally it would appear the setup of inbuilt drives is done in the wrong location:
* These lines: https://github.com/Badgerati/Pode/blob/317cc64fff2b217b358b2339824ff0399ad5800d/src/Private/Server.ps1#L49-L50
* Should actually be after this line: https://github.com/Badgerati/Pode/blob/317cc64fff2b217b358b2339824ff0399ad5800d/src/Private/Server.ps1#L71

In your ticket as well, when you call `Set-PodeDefaultFolder`, you can omit the `Get-PodeServerPath` - this is done for you in the other file when Pode interally called `Join-PodeServerRoot` 😄:

```powershell
Set-PodeDefaultFolder -Type Errors -Path 'ErrorViews'
```

Badgerati (changes requested):

src/Public/Core.ps1
CedX: Moving `Add-PodePSInbuiltDrive` after `$null = [...]` => It works!

Removing `Get-PodeServerPath` in the call to `Set-PodeDefaultFolder` => Error (i.e. "Path not found").
`Set-PodeDefaultFolder` immediately calls `Test-Path`...

Badgerati: Ooh, good point. It seems `Set-PodeDefaultFolder` is missing the following just before the `Test-Path`:

```powershell
$Path = Join-PodeServerRoot $Path
```

CedX: Fixed 😊

Badgerati (approved):
LGTM

Badgerati:
I've flagged this for 2.14.0, but I'll see if I can do a 2.13.3 release this week.

CedX:
All that remains for me is to say a big thank you.
Thank you @Badgerati for your responsiveness, and of course, thank you for Pode. You've done an exceptional job with this library.

PR #1756, opened 2026-08-03
Enhances State management to be thread-safe; Enhances State management to Save and Restore true object types

Badgerati:
### Description of the Change
State management has always been a standard Hashtable, which isn't thread-safe - and it's recommended to use `Lock-PodeObject`.

This PR converts the State to a ConcurrentDictionary, and adds helper functions for creating other concurrent/thread-safe data structures; such as lists, bags, set, etc.

Set, Get, Save, and Restore are all now more thread-safe - reducing the need for `Lock-PodeObject`. Furthermore, any prior states restored will be automatically converted into a ConcurrentDictionary.

The State's saved JSON object has been expired, to save the actual type being saved - so the true type is restored.


Badgerati:
Closes #1473

PR #1765, opened 2026-08-09
Fix to accept repeated HTTP request header fields

### LINKED ISSUE #1764: Pode returns HTTP 400 for repeated request header fields
### Describe the Bug
Pode returns HTTP 400 before route or middleware processing when an HTTP request contains two field lines with the same name.

The HTTP listener stores request headers in a case-insensitive `Hashtable` and calls `Headers.Add(h_name, h_value)` for each field line. `Hashtable.Add` throws `ArgumentException` when the key already exists. The listener converts this parsing exception to HTTP 400.

This affects valid list-based fields that can occur on more than one line. For example, a RingCentral webhook validation request sends:

```http
Accept-Encoding: gzip, deflate
Accept-Encoding: UTF-8
```

The same values work when the sender combines them into one field line:

```http
Accept-Encoding: gzip, deflate, UTF-8
```

### Steps To Reproduce
1. Start a Pode HTTP server with a basic route:

   ```powershell
   Start-PodeServer {
       Add-PodeEndpoint -Address localhost -Port 8080 -Protocol Http
       Add-PodeRoute -Method Get -Path '/headers' -ScriptBlock {
           Write-PodeJsonResponse -Value @{
               AcceptEncoding = Get-PodeHeader -Name 'Accept-Encoding'
           }
       }
   }
   ```

2. Send a raw HTTP request that contains two `Accept-Encoding` field lines:

   ```http
   GET /headers HTTP/1.1
   Host: localhost:8080
   Accept-Encoding: gzip, deflate
   Accept-Encoding: UTF-8
   Connection: close

   ```

3. Pode returns `HTTP/1.1 400 BadRequest` and the route does not run.

The script below demonstrates the issue:
1. Run `pwsh -NoProfile -File '.\duplicate-request-headers-demo.ps1' -Server` 
2. In a separate terminal, run `pwsh -NoProfile -File "D:\Pode\duplicate-request-headers-demo.ps1" -Client -Port 8080` 

```powershell
# duplicate-request-headers-demo.ps1

[CmdletBinding(DefaultParameterSetName = 'Server')]
param(
    [Parameter(ParameterSetName = 'Server')]
    [switch]
    $Server,

    [Parameter(ParameterSetName = 'Client')]
    [switch]
    $Client,

    [int]
    $Port = 8080,

    # To point the script do a different build of Pode, such as one where the issue is fixed.
    [Parameter(ParameterSetName = 'Server')]
    [string]
    $PodeModulePath
)

Set-StrictMode -Version Latest
$ErrorActionPreference = 'Stop'

if ($PSCmdlet.ParameterSetName -eq 'Client') {
    # Same request as the issue: two Accept-Encoding field lines.
    $request = @(
        'GET /headers HTTP/1.1'
        "Host: localhost:$Port"
        'Accept-Encoding: gzip, deflate'
        'Accept-Encoding: UTF-8'
        'Connection: close'
        ''
        ''
    ) -join "`r`n"

    Write-Output 'Sending:'
    Write-Output $request

    $tcp = [System.Net.Sockets.TcpClient]::new('127.0.0.1', $Port)
    try {
        $stream = $tcp.GetStream()
        $bytes = [System.Text.Encoding]::ASCII.GetBytes($request)
        $stream.Write($bytes, 0, $bytes.Length)
        $stream.Flush()

        # Installed Pode can leave the socket open after HTTP 400.
        $stream.ReadTimeout = 2000
        $buffer = [byte[]]::new(4096)
        $memory = [System.IO.MemoryStream]::new()
        try {
            while ($true) {
                try {
                    $read = $stream.Read($buffer, 0, $buffer.Length)
                    if ($read -le 0) {
                        break
                    }
                    $memory.Write($buffer, 0, $read)
                } catch [System.IO.IOException] {
                    break
                }
            }

            $response = [System.Text.Encoding]::ASCII.GetString($memory.ToArray())
        } finally {
            $memory.Dispose()
        }
    } finally {
        $tcp.Dispose()
    }

    $statusLine = ($response -split "`r`n", 2)[0]
    Write-Output ''
    Write-Output 'Response:'
    Write-Output $response
    Write-Output ''

    if ($statusLine -match '\s400\s') {
        Write-Output '[REPRODUCED] HTTP 400 - repeated Accept-Encoding fields failed before the route.'
    } elseif ($statusLine -match '\s200\s') {
        Write-Output '[FIXED] HTTP 200 - route ran and returned Accept-Encoding.'
    } else {
        Write-Output "[UNEXPECTED] $statusLine"
    }

    return
}

if ([string]::IsNullOrWhiteSpace($PodeModulePath)) {
    Import-Module -Name Pode -MaximumVersion 2.99.99 -ErrorAction Stop
} else {
    Import-Module -Name (Resolve-Path -LiteralPath $PodeModulePath).Path -Force -ErrorAction Stop
}

Write-Output "Listening on http://localhost:$Port/"
Write-Output 'In a second terminal, run:'
Write-Output "  pwsh -NoProfile -File `"$PSCommandPath`" -Client -Port $Port"
Write-Output 'Press Ctrl+C to stop the server.'
Write-Output ''

$env:PODE_DEMO_PORT = "$Port"
Start-PodeServer {
    $listenPort = [int]$env:PODE_DEMO_PORT
    Add-PodeEndpoint -Address localhost -Port $listenPort -Protocol Http
    Add-PodeRoute -Method Get -Path '/headers' -ScriptBlock {
        Write-PodeJsonResponse -Value @{
            AcceptEncoding = Get-PodeHeader -Name 'Accept-Encoding'
        }
    }
}
```

### Expected Behavior
Pode should accept repeated list-based request fields and preserve their values in received order. For `Accept-Encoding`, `Get-PodeHeader` should return the combined value:

```text
gzip, deflate, UTF-8
```

The request should reach the route and return HTTP 200.

### Screenshots
Screenshot of demo script provided:

Image

### Platform
 - OS: Windows 11
 - Browser: Not applicable; reproduced with a raw TCP HTTP request
 - Versions: 
   - Pode: v2.13.4
   - PowerShell: PS7.6.4

### Additional Context
The failure occurs in:

```text
src/Listener/Protocols/Http/PodeHttpRequestStrategy.cs
```

Specifically, `ParseHeaders` calls:

```csharp
Headers.Add(h_name, h_value);
```

The parser fails before Pode creates a usable web event, so custom middleware cannot handle or normalize the repeated fields.

SsSonic3:
### Description of the Change
HTTP request header parsing now accepts repeated field lines with the same name. When a header name already exists, Pode appends the new value with a comma instead of calling `Hashtable.Add`, which previously threw and returned HTTP 400 before any route or middleware ran.

This matches common list-based header behavior and unblocks clients such as RingCentral webhooks that send:

```http
Accept-Encoding: gzip, deflate
Accept-Encoding: UTF-8
```

### Related Issue
Closes https://github.com/Badgerati/Pode/issues/1764

### Examples
Before this change, a request with two `Accept-Encoding` lines returned HTTP 400 and never reached the route.

After this change:

```powershell
Start-PodeServer {
    Add-PodeEndpoint -Address localhost -Port 8080 -Protocol Http
    Add-PodeRoute -Method Get -Path '/headers' -ScriptBlock {
        Write-PodeJsonResponse -Value @{
            AcceptEncoding = Get-PodeHeader -Name 'Accept-Encoding'
        }
    }
}
```

A raw request with both `Accept-Encoding` lines returns HTTP 200 and:

```json
{ "AcceptEncoding": "gzip, deflate, UTF-8" }
```

### Additional Context
- Change is in `src/Listener/Protocols/Http/PodeHttpRequestStrategy.cs` (`ParseHeaders`).
- Added an integration test in `tests/integration/RestApi.Tests.ps1` that sends two `Accept-Encoding` field lines over TCP and asserts HTTP 200 plus the combined header value.
- Integration tests were run locally against this change (`86` passed).

### DIFF
--- a/src/Listener/Protocols/Http/PodeHttpRequestStrategy.cs
+++ b/src/Listener/Protocols/Http/PodeHttpRequestStrategy.cs
@@ -309,7 +309,14 @@ namespace Pode.Protocols.Http
                 {
                     h_name = h_line.Substring(0, h_index).Trim();
                     h_value = h_line.Substring(h_index + 1).Trim();
-                    Headers.Add(h_name, h_value);
+                    if (Headers.ContainsKey(h_name))
+                    {
+                        Headers[h_name] = $"{Headers[h_name]}, {h_value}";
+                    }
+                    else
+                    {
+                        Headers.Add(h_name, h_value);
+                    }
                 }
             }
 
--- a/tests/integration/RestApi.Tests.ps1
+++ b/tests/integration/RestApi.Tests.ps1
@@ -27,6 +27,12 @@ Describe 'REST API Requests' {
                     Write-PodeJsonResponse -Value @{ Result = 'Pong' }
                 }
 
+                Add-PodeRoute -Method Get -Path '/headers' -ScriptBlock {
+                    Write-PodeJsonResponse -Value @{
+                        AcceptEncoding = Get-PodeHeader -Name 'Accept-Encoding'
+                    }
+                }
+
                 function Write-InnerImportedResponse {
                     Write-PodeJsonResponse -Value @{ Message = 'Inner Hello' }
                 }
@@ -122,6 +128,38 @@ Describe 'REST API Requests' {
     It 'responds back with pong' {
         $result = Invoke-RestMethod -Uri "$($Endpoint)/ping" -Method Get
         $result.Result | Should -Be 'Pong'
+    }
+
+    It 'combines duplicate request header values' {
+        $client = [System.Net.Sockets.TcpClient]::new('127.0.0.1', $Port)
+        try {
+            $stream = $client.GetStream()
+            $request = @(
+                'GET /headers HTTP/1.1'
+                "Host: localhost:$($Port)"
+                'Accept-Encoding: gzip, deflate'
+                'Accept-Encoding: UTF-8'
+                'Connection: close'
+                ''
+                ''
+            ) -join "`r`n"
+
+            $requestBytes = [System.Text.Encoding]::ASCII.GetBytes($request)
+            $stream.Write($requestBytes, 0, $requestBytes.Length)
+            $stream.Flush()
+
+            $reader = [System.IO.StreamReader]::new($stream)
+            $response = $reader.ReadToEnd()
+        }
+        finally {
+            $client.Dispose()
+        }
+
+        $responseParts = $response -split "`r`n`r`n", 2
+        $responseParts[0] | Should -Match '^HTTP/1.1 200'
+
+        $body = $responseParts[1] | ConvertFrom-Json
+        $body.AcceptEncoding | Should -Be 'gzip, deflate, UTF-8'
     }
 
     It 'responds back with 404 for invalid route' {


Badgerati (approved):
LGTM, thanks! 😄

